Issue #223, opened 2014-11-19
114th Congress Update Strategy


JoshData:
Hey,

Let's use this issue to plan how to update for the 114th Congress.

Something like:
- [x] Identify current Members that won, copy their term info forward but update dates. Check if anyone changed district/party (not likely). Be careful if any representatives became senators. - @joshdata
- [x] Move incumbents that lost to the historical file. (nb. Not everyone was up for reelection.) - @joshdata
- [x] Identify if any new members served in Congress previously and bring their records forward w/ a new term. - @joshdata
- [x] Add remaining new members.
- [x] Review junior/senior status
- [x] Add end dates to all current `leadership_roles` (no leadership_role should extend into the next Congress) - @rshorey or @handlers 
- [x] Add `leadership_roles` for the 114th Congress (#235)
- [ ] Not sure what to do about committees. There won't be committee assignments initially.
- [ ] [Photos](https://github.com/unitedstates/images)? Did/will GPO publish a new member directory?
- [X] Fill in bioguide IDs when they become available.
- [X] Delete retirees' `social-media` records.
- [ ] Add new members' social media accounts. (#234)
- [ ] After the 114th Congress begins, fill in missing data.

I'd like to set up a google doc spreadsheet where the rows are all of the offices that were up for election (all representatives, presumably all delegates, not sure about Puerto Rico, and one of the Senate classes) so we can make sure we take care of each election result.

@dwillis Any overlap with open elections data?

Did I miss anything?

cc @handlers @konklone

JoshData:
All of the core stuff is ready to go in #230. I want to get this merged asap, so last call.... Thanks for everyone's help getting this far.

handlers:
Oh wow — thanks! :+1:  from me on the merge of #230 .

rshorey:
Sounds good to me. Do we have the promised xml file with bioguide IDs yet?

JoshData:
A little birdie told me to guess the IDs by looking at the next sequential IDs --- I actually committed a script that reports the next set of bioguide IDs. (It doesn't work perfectly.) So I did that and entered all of the bioguide IDs for the new members. (I also copied in the new opensecrets IDs.)

We should be on the lookout for when LIS IDs are posted at http://www.senate.gov/general/contact_information/senators_cfm.xml.

rshorey:
[...]

PR #230, opened 2015-01-03
114th Congress updates

JoshData:
See #223.


konklone:
The tests are failing because the `114th_congress.py` script loads a CSV that's not in version control. But the root issue is the script shouldn't execute simply by being imported (which is all the test script does). 

When I added the `workout` script, I also adjusted the other scripts to only execute when run directly. This means putting the bulk of the script inside of a function, e.g. `run()`, and then putting this at the bottom of the script:

```python
if __name__ == '__main__':
  run()
```

JoshData:
It *is* in version control, actually. But thanks for fixing the script to not run. Travis still shows a failure.

konklone:
There were two more scripts that needed a `run()` method to protect against execution-on-import. Added that, I suspect this Travis build will pass.

konklone:
OK, build is fixed. It did throw a few pyflakes warnings (we add that in as an `after_success` command for Travis to run, but don't break the build on its failure) so I removed some unused imports and variables in cc0da3b. Build restored to pristine state. ;)

PR #231, opened 2015-01-06
Add Bradley Walker and move Rep. Coble to legislators-historical.yaml

handlers:
Looks like something was amiss in the new_leg spreadsheet (Coble's bioguide id had been in the new_id column), so we added an extra term for Rep. Coble, who retired. He's replaced by [Bradley Walker](http://bioguide.congress.gov/scripts/biodisplay.pl?index=W000819) in the NC-6.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -38683,3 +38683,4 @@
     end: '2017-01-03'
     state: NC
     district: 6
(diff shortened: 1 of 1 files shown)


JoshData:
Looks good. I'll let you do the honors.

PR #233, opened 2015-01-08
Thomas IDs added to new reps where available

rshorey:
note that new senators aren't finished here and will have to be done by hand.


JoshData:
I can do the senate IDs if no one is working on it.

rshorey:
I'm definitely not going to get to it today, so go for it!

JoshData:
Okay, committed in 769c0d5ffe48c779d3c61bb268d97adf67b30116. Everyone has a thomas ID now.

PR #238, opened 2015-02-02
added some missing fax numbers

crdunwel:


JoshData:
+1

handlers:
So it's worth noting that a couple of these aren't sourced from elected officials' official websites. As @crdunwel notes, there's nothing in the [README](https://github.com/unitedstates/congress-legislators#legislators-data-dictionary) specifying where the fax numbers should come from. Is this something we want to specify?

JoshData:
I don't think the source matters as long as it's accurate. But for the sake of provenance, it's always nice to have a source, or who is vouching for the accuracy of the information, in the commit message.

crdunwel:
Just google the fax numbers and you'll find bunch of site referencing them.

@JoshData - @handlers has a point. The government may be wrong sometimes but in general government sites are the most "official" source. Other sources may just be secondhand, blind leading the blind, or just outdated.

My feeling is that until proven otherwise, it's better to use the last widely listed or known fax number than to have nothing at all.

handlers:
OK, I am good with merging this now that we've clarified.

crdunwel:
@handlers merge?

PR #246, opened 2015-02-10
Update contact_form link

crdunwel:
Does this repo grab data from the contact-congress repo for contact links? That repo is perhaps the most up to date for contact form links.


JoshData:
If the other repo is being maintained better, we can drop the field from this one maybe?

PR #250, opened 2015-02-23
Most Committees for 114th Congress

handlers:
This is the output of `committee_membership.py`, run with `scrape_house_alt()` instead of the deprecated `scrape_house()`. I spot checked quite a few of the committee and subcommittee member lists, and they looked pretty good, with the following exceptions:

* House Committee on Natural Resources (`HSII`) subcommittees only have ex officio members.
* Senate Committee on Environment and Public Works (`SSEV`) subcommittees still have no members.
* **Some committees may have outdated House members**, since the House [doesn't publish joint committee members](https://github.com/unitedstates/congress-legislators/blob/master/scripts/committee_membership.py#L285). I manually cleaned up Commission on Security and Cooperation in Europe (`JCSE`) ([data source](http://www.csce.gov/index.cfm?FuseAction=AboutCommission.WorkOfCommission)), Joint Committee on the Library (`JSLC`) ([data source](http://cha.house.gov/jointcommittees/joint-committee-library)), and the Joint Committee on Taxation (`JSTX`) ([data source](https://www.jct.gov/)). The Joint Economic Committee's website's roster has [not been updated for the 114th](http://www.jec.senate.gov/public/index.cfm?p=HouseMembers), nor has the [Joint Committee on Printing's](http://cha.house.gov/jointcommittees/joint-committee-on-printing).

Would love to get a sanity check on this, but it seems OK to me!

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -10641,12 +10641,12 @@ JSTX:
   bioguide: S000770
   thomas: '01531'
   chamber: senate
-- name: Dave Camp
+- name: Paul Ryan
   party: majority
   rank: 1
   title: Chairman
-  bioguide: C000071
-  thomas: '00166'
+  bioguide: R000570
+  thomas: '01560'
   chamber: house
 - name: Sam Johnson
   party: majority
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -10346,15 +10346,9 @@ JCSE:
   bioguide: A000055
   thomas: '01460'
   chamber: house
-- name: Phil Gingrey
-  party: majority
-  rank: 4
-  bioguide: G000550
-  thomas: '01720'
-  chamber: house
 - name: Michael C. Burgess
   party: majority
-  rank: 5
+  rank: 4
   bioguide: B001248
   thomas: '01751'
   chamber: house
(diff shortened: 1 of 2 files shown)


JoshData:
+1 (I didn't really look over anything carefully, but it seems great!)

rshorey:
+1

JoshData:
In JSEC, John Campbell (01816) is still in there but he's no longer serving. Mind if I just remove entries like that, while we wait for other info to update the derelict committees?

handlers:
Yeah, that sounds good to me, @JoshData.

JoshData:
One of my users spotted a problem. The data on House Oversight doesn't match the House Oversight webpage:

http://clerk.house.gov/committee_info/index.aspx?comcode=GO00 => chair is John Mica
http://oversight.house.gov/subcommittee/full-committee/ => chair is Jason Chaffetz

handlers:
Thanks for the heads, @JoshData. I just spoke with someone from the committee on the phone, and apparently "both are correct" — Mica has been on the committee the longest, while Jason Chaffetz is currently the chair. I'm going to check to make sure that the rest of the memberships align, but expect a PR soon.

JoshData:
Interesting.

handlers:
So the house.gov site lists three members that are not on the committee. I called the committee back and confirmed that the following members are not on it (they do not appear on the House clerk's page):

* Davis
* Speier
* Cardenas

handlers:
see #256

PR #256, opened 2015-02-26
Make Chaffetz chair of HSGO, ref pr-250

handlers:


JoshData:
Possibly keep the rank/order but move `title` down? Sounds like Chaffetz has low rank but somehow is the chair.

dwillis:
Since the House GOP has term limits for committee chairs, more and more of them will be people who don't have the most seniority on a committee. At NYT, I handle this by adding a `chair` boolean attribute to the data, independent of seniority or rank in party.

handlers:
PR updated along the lines of what @JoshData suggested. 

@dwillis I think that makes sense, but for my purposes, I think `title: chair` serves as that boolean.

JoshData:
+1

PR #259, opened 2015-03-09
normalize religion value for Robert Wittman

wilsaj:
to match other legislators of the same religious affiliation (Episcopal -> Episcopalian)


JoshData:
Merged. But we only have this data in for legislators who were serving the small time window when I cared about this field (2004-2008ish), so we should consider deleting the field entirely...

Issue #267, opened 2015-04-10
bob menendez website url is wrong


timball:
i'm submitting a PR to fix this commit: 8d14459d4343f3fd2521f3ff24ae41743eebdac2

--timball

PR #268, opened 2015-04-10
senator menendez senate url has a www

timball:
this fixes issue #267 

--timball


handlers:
closed in 696d3807ff8703c205e059cec90ea56cff4ad299. didn't need to update historical urls.

PR #269, opened 2015-04-10
www. for menendez' website

handlers:
h/t @timball


crdunwel:
why don't we make it https while we're at it?

PR #271, opened 2015-04-15
new contact_form fields

crdunwel:


handlers:
@crdunwel some of these are the unstyled iframes containing members' webforms. Are you counting on those urls for a specific implementation or can I merge then amend the `forms.house.gov` forms to point to styled forms?

crdunwel:
@handlers I'm lazy so I just wrote a script to copy them from https://github.com/unitedstates/contact-congress . If the spec calls for the styled form url then feel free to change them as appropriate. :+1:

JoshData:
What's the status of this PR?

PR #272, opened 2015-04-15
Now 114th Congress

martinburch:
Noticed this constant is outdated. Should this be updated?


martinburch:
Is the 114 Congress not ready yet, as evidenced by https://github.com/unitedstates/congress-legislators/blob/master/scripts/114th_congress.py or can that file be removed now?

JoshData:
The script can be removed.

The constant can be updated, although it's only used in one place.

timball:
issue #280 is related to this .

Issue #278, opened 2015-05-14
senate urls should have www prefix


timball:
i know people hate on the "www." in a url but NOT the United States Senate. Therefore the stuff in "committees-current.yaml" is mostly wrong wrt to senate urls. ... PR coming shortly .

--timball

timball:
PR fixes this .

PR #279, opened 2015-05-14
fixes issue #278

timball:
senate urls prefer the "www." and many hostnames will not redirect to canonical name w/ the www.

--timball

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -38118,8 +38118,8 @@
     state: LA
     district: 6
     party: Republican
-    address: 204 Longworth HOB; Washington DC 20515-1806
-    office: 204 Longworth House Office Building
+    address: 204 Cannon HOB; Washington DC 20515-1806
+    office: 204 Cannon House Office Building
     phone: 202-225-3901
     url: https://garretgraves.house.gov
     fax: 202-225-7313
@@ -38852,7 +38852,7 @@
     state: VI
     district: 0
     party: Democrat
-    address: 509 Cannon HOB; Washington DC 20515-5501
+    address: 509 Cannon HOB; Washington DC 20515-5500
     office: 509 Cannon House Office Building
     phone: 202-225-1790
     url: https://plaskett.house.gov
@@ -38971,7 +38971,7 @@
   name:
     first: Aumua
     last: Amata
-    official_full: Amata Coleman Radewagen
+    official_full: Aumua Amata Coleman Radewagen
   bio:
     gender: F
     birthday: '1947-12-29'
@@ -38982,7 +38982,7 @@
     state: AS
     district: 0
     party: Republican
-    address: 1339 Longworth HOB; Washington DC 20515-5201
+    address: 1339 Longworth HOB; Washington DC 20515-5200
     office: 1339 Longworth House Office Building
     phone: 202-225-8577
     url: https://radewagen.house.gov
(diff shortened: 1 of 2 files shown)


konklone:
:+1:

JoshData:
I'm not sure it makes sense to update the historical webpage URLs in legislators-current.yaml. Otherwise +1.

konklone:
Mmmm, good catch. In fact, it's better not to, since their only real value is showing what the URL was at that time. So retroactively updating them is all (though perhaps minor) downside.

@timball, are any of these updates to current terms? If so, can you update the PR to limit it to just current terms?

timball:
the frak are you two talking about ? this is to committees-current.yaml . so ... currently the senate needs www. in the urls . merge this . pedantic whose-a-whats-itz . did not touch any of the "historical" because honestly don't even understand the need for historical thingies to exist . that's why we use git right ? track file changes and evolutions . buttons . magnets . balls . 

--timball

timball:
oh and another things IF the historic files DIDN'T contain the www. then those senate urls would be WRONG . so yeah . they frakking need it . 

--timball

JoshData:
Talkin' 'bout `legislators-current.yaml`.

timball:
OH! this about the "1995" comment . yes . that was irony . used it to reference the prince song "party like it's 1999" . but in web terms , www. feels 1995 to me . regardless all my statements still stand . if the purpose of this repo is to have accurate data then all senate urls need the www. 

--timball

timball:
or maybe my menendez pull .. yeah his thing needs www. too otherwise it's wrong . don't know . don't care . this is what i get for trying to help .

legislators-current.yaml
```diff
@@ -817,7 +817,7 @@
     state: NJ
     party: Democrat
     class: 1
-    url: http://menendez.senate.gov
+    url: http://www.menendez.senate.gov
```
konklone: This edit is good. The other 3 changes above in this file should be reverted. I can submit a PR to the PR.

konklone:
@timball, we've asked lots of people to edit their PRs before merging, it's a normal thing.

The non-current term info for Menendez should keep the (possibly broken) URLs that we logged for them at the time. I'll submit a PR to your PR on your repo, which you can merge and then we'll merge this one.

timball:
it's dumb to hold onto inaccurate and poorly verified bad data . there was never a time that a bare "menendez.senate.gov" worked . and in the weird chance that it did it would have redirected to the www. do whatever you're gonna do but it's dumb . either way merge the other things because accurate is what i'm going for . 

my suspicion is that whomever put that data in there had the hipster www. angst . and just didn't want to see www. dirty their beautifully broken data . that person is bad and should be spanked . this is about accurate data , not beautiful data . EVEN if the the thing redirects you i contend that 'url' should be the url that the user gets too not the one that redirects you away (which btw my PR doesn't actually do the right thing i just noticed that http://www.rules.senate.gov actually redirects you to http://www.rules.senate.gov/public/ so whatever . my whole thing is moot dog shit bite me ) . but even if you go and fix this shit then there will be accurate shit . 

PAX PER IMPERIUM

--timball

konklone:
I filed https://github.com/timball/congress-legislators/pull/1 for you to merge, @timball, which will update this PR in place and clear it for our merge.

konklone:
Thanks for accepting the PR, and for the fixes!

JoshData:
My patience for nonsense ran out. I've locked the thread. Point taken. In the future, when blaming anonymous people, it's worthwhile to run a `git blame` and look up why the change was made in the first place. (In this case, the Menendez URLs date back to my initial commit to this repository.)

PR #290, opened 2015-05-21
Joe is his nickname

timball:
Joe Crowley has a nickname , it's Joe . 

--timball


konklone:
Okay, http://crowley.house.gov/about-me/full-biography is good enough for me!

Issue #286, opened 2015-05-17
house leadership offices


timball:
house leadership offices are separate from house member offices and consequently have their own staff and associated website . proposal is to create `house-leadership.yaml` to capture this information . a complementary `senate-leadership.yaml` might be warranted if they had websites , but they don't yet . 

example entities to include :
* Speaker
* Majority Leader
* Majority Whip
* Republican Conference Chairman
* Republican Policy Committee Chairman
* Democratic Leader
* Democratic Whip
* Assistant Democratic Leader
* Democratic Caucus Chairman

--timball

PR #291, opened 2015-05-21
added leadership file

timball:
initial revision of a leadership file w/ information not captured anywhere else addresses issue #286

--timball

### DIFF
--- a/leadership.yaml
+++ b/leadership.yaml
@@ -58,7 +58,7 @@
   bioguide: C001038
   url: http://www.dems.gov/
 
--type: senate
+- type: senate
   title: Majority Leader
   office: Senate Majority Leader
   bioguide: M000355
--- a/leadership.yaml
+++ b/leadership.yaml
@@ -1,0 +1,119 @@
+- type: house
+  title: Speaker the House
+  office: Speaker
+  bioguide: B000589
+  url: http://speaker.house.gov/
+
+- type: house
+  title: Majority Leader
+  office: Majority Leader
+  bioguide: M001165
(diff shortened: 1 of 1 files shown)


joelcollinsdc:
The official website for the speaker is www.speaker.gov

timball:
huh .. so it is . changes coming real soon

leadership.yaml
```diff
@@ -0,0 +1,119 @@
+- type: house
+  title: Speaker the House
+  office: Speaker
+  bioguide: B000589
```
konklone: I'd love others' opinions using the data, but I think it's better to keep the bioguide out of this file, and rely on the identification of the leader in `legislators-current.yaml`. I say that partly over redundancy, but mostly without the `bioguide` (and my impending comment about URLs), this file can stay mostly static, instead of needing to change with shifts in people.

leadership.yaml
```diff
@@ -0,0 +1,119 @@
...
+  title: Republican Conference Vice Chair
+  office: Republican Conference
+  bioguide: B000575
+  url: http://www.republican.senate.gov/
+
+- type: senate
+  title: Minority Leader
+  office: Senate Minority Leader
+  bioguide: R000146
+  url: http://www.reid.senate.gov
```
konklone: This URL isn't the "URL of the office of Senate Minority Leader", it's Reid's individual MoC site, so I think is not correct for this field. Keeping the URLs to their respective institutions also keeps this file more static.

JoshData:
+1 to @konklone's comments here too.

PR #294, opened 2015-06-15
added FEC IDs for three historical legislators

mrumsky:


JoshData:
Looks good, merging, thanks!

PR #298, opened 2015-07-11
Remove .python-version

konklone:
And expect users of the repo to gitignore it and use their own.


JoshData:
What was this file for?

konklone:
This file lets `pyenv` users set what version of Python should be automatically made active upon `cd`-ing into the given directory. It was versioned as a convenience, and to sync up the version of Python we test this project out on with what it tells developers to use.

It's just tough for it to keep up, since we don't test the app fully on every minor version upgrade. Seems better managed in the README. Open to other ideas!

JoshData:
> Open to other ideas!

I didnt even know we had this file in the first place, so this is all +1s from here!

Issue #293, opened 2015-05-28
Social media urls to add to your legislators data set


pardery:
I work with the Library of Congress, and I am using your 'legislators-current' and 'legislators-social-media' yaml files to facilitate scoping for our US Congress web archive collection. Thanks for the data!

As it turns out, you had a good chunk of social media sites that we were missing. However, we also seem to have a descent amount of data that you are missing. I have created the following json file outlining the social media accounts that we have found, but are missing from your data set, per Bioguide ID.  @dwillis [informed me] (https://github.com/unitedstates/unitedstates.github.io/issues/4) that some of it may not be pertinent to the list you keep, but I thought I might as well pass it along:

https://gist.github.com/pardery/84fc56c836c4b1f02708

Thanks,
Phil

gmj2053:
Re: "official".  We only scope in sites that are linked from the congressional website as part of their social media framework.  Where does that come from for your set?
(Phil and I work together)

konklone:
So if these are all linked to from their congressional websites, that's often a strong enough signal to accept those handles, but review is needed.

There are a few that we exclude anyway, when they're not actually for the member, like: `http://www.flickr.com/photos/republicanconference/` or `http://www.youtube.com/HouseConference`. Those are often used as placeholders until the MoC gets their own account.

Also, not all of the Twitter accounts are legit. `http://twitter.com/SenRandPaul` is listed, but that account doesn't exist. `http://twitter.com/alangrayson` is also listed, but that doesn't look legit, and http://grayson.house.gov doesn't link to it (or to any Twitter account). `https://twitter.com/sethmoulton` works, and is linked to by his [official site](http://grayson.house.gov/), but the Twitter bio links to Moulton's campaign site, which is a mistake on _someone's_ part and potentially opens up Moulton to charges that he's using official resources to benefit his political campaign. I'd at least want to call Moulton's office to confirm that the account is meant to be his official legislative account.

We also haven't (yet) been tracking Flickr accounts. It looks like you have 304 listed in that gist (not counting the `/republicanconference/` account). That's more widespread than I thought -- perhaps the project should consider tracking Flickr accounts too.

We don't track G+ accounts, and there are only 33 `plus.google.com` URLs in that gist -- plus G+ is a wasteland -- so that doesn't seem worth tracking. We're also not likely to incorporate a Picasa account (e.g.  http://picasaweb.google.com/congresswomanpingree/)

[...]

PR #299, opened 2015-07-11
Twitter and Facebook account update

konklone:
This updates Twitter and Facebook accounts based on a current sweep. I also reviewed the accounts submitted in #293. We picked up all but 1 of the legit accounts in that submission through our automated scan, and that 1 led to a bug-fix in https://github.com/unitedstates/congress-legislators/commit/8d0a076e885daed028725bad2e14c698e8076942 to handle protocol-relative URLs to `twitter.com`.

I hand-verified all Twitter and Facebook accounts as cross-linked between the account and the official MoC site. One exception is `@teammoulton` on Twitter, which is the official account of Rep. Moulton, even though his official site links to `@sethmoulton`.

I'm tapped out for now on social media updates -- there's still some work to do on YouTube accounts and Instagram accounts.


konklone:
Hearing no objection...

PR #301, opened 2015-07-19
Don't include trailing paths in official members' website URLs

### LINKED ISSUE #300: Make sure House/Senate websites use the domain root
Instead of what they redirect to. @gmj2053 has identified 1 to 3 examples (including Crenshaw) of us using a particular index-y URL as the homepage for a member, which can break.

konklone:
House MoC URLs used to occasionally be subpaths, in earlier times, but today they're always subdomains. We can safely omit everything after the `.gov`.

We still don't naively trust the official directory, and follow redirects from whatever domain is given. This has the effect of occasionally fixing out-of-date domains (e.g. the directory lists `randyforbes.house.gov` which redirects to `forbes.house.gov`). 

This also removes some `www` subdomains and moves some URLs to `https://` (though these could also be a function of just not having run this script in a while).

Fixes #300. Thanks to @gmj2053 for reporting this and motivating the fix!

### DIFF
--- a/scripts/house_websites.py
+++ b/scripts/house_websites.py
@@ -37,7 +37,7 @@ def run():
         states.append(state)
 
   destination = "legislators/house.html"
-  url = "http://house.gov/representatives/"
+  url = "http://www.house.gov/representatives/"
   body = utils.download(url, destination, force)
   if not body:
     print("Couldn't download House listing!")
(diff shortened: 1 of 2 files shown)


scripts/house_websites.py
```diff
@@ -69,18 +69,26 @@ def run():
...
-      # kill trailing slashes
-      url = re.sub("/$", "", url)
+      # kill everything after the domain
+      url = re.sub(".gov/.*$", ".gov", url)
 
       if state == "AQ":
         state = "AS"
       full_district = "%s%02d" % (state, int(district))
       if full_district in by_district:
+        print("[%s] %s" % (full_district, url))
```
konklone: There's extra debug output now -- seems harmless, and helpful.

scripts/house_websites.py
```diff
@@ -37,7 +37,7 @@ def run():
         states.append(state)
 
   destination = "legislators/house.html"
-  url = "http://house.gov/representatives/"
+  url = "http://www.house.gov/representatives/"
```
konklone: I guess the House moved to a `www` canonical domain for its main site? I fixed this to eliminate the need to follow a redirect.

PR #302, opened 2015-07-30
Update legislators-social-media.yaml

hpetru:
some new twitter usernames


konklone:
@hpetru, thank you very much for contributing here! However, these mostly look like campaign accounts. We only track official accounts, and generally want to see them linked from the member's official House or Senate website.

`repmccaul` looks like an accurate update for an official account. If you want to update your pull request to limit it to that change, along with any others that you've verified are linked from their official .gov homepages, we'd be happy to :eyes: and merge.

PR #303, opened 2015-08-02
Add Twitter account IDs to legislators-social-media.yaml

dannguyen:
Since we're adding Facebook/Instagram/Youtube ID numbers for the legislators, might as well add Twitter ID numbers, which are helpful when screen_names change as they sometimes do.

This:

~~~yaml
- id:
    bioguide: G000386
    thomas: '00457'
    govtrack: 300048
  social:
    twitter: ChuckGrassley
    facebook: grassley
    youtube: senchuckgrassley
    facebook_id: '106480645796'
    youtube_id: UCiCPRkcLMInpDeOqGkMTt4w
~~~

Becomes:

~~~yaml
- id:
    bioguide: G000386
    thomas: '00457'
    govtrack: 300048
  social:
    twitter: ChuckGrassley
    facebook: grassley
    youtube: senchuckgrassley
    facebook_id: '106480645796'
    youtube_id: UCiCPRkcLMInpDeOqGkMTt4w
    twitter_id: 10615232
~~~

This script also removes invalid Twitter names if they aren't returned by the lookup call in the API. In this run, 3 invalid names were found and removed from the YAML: OlsonPressShop, McCaulPressShop, USRepRussell.
A quickie (Python 3) script is added to show how the reconciliation and batch insertion was done: [scripts/fetch_and_insert_twitter_ids.py](scripts/fetch_and_insert_twitter_ids.py)


dwillis:
I've got no issues with this. +1

nickom:
Very smart addition, chasing changing Twitter handles is a hassle +1

konklone:
This is a smart addition and I also have no problems with it. I also don't mind that it's standalone and not integrated with `social_media.py`. 

However, I would like to see if we can't make the script integrate properly. 

* Right now, it's all commented out right now, so as not to break the Travis build. That looks to be fixable just by adding `tweepy` to `scripts/requirements.txt`, which Travis is set to use to install dependencies. I was able to run the Travis script successfully on my machine after doing this and installing `tweepy`.
* The output data path is set to `/tmp/new-social-media.yaml`. The repo already allows scripts to make/use a `scripts/cache` directory for temporary work. That said, I think you could just have it overwrite the social media YAML file automatically, and only commit the file after verification.
* The twitter credentials path is hardcoded to `expanduser("~/.creds/twit.json")`. This would be a better place to use the `scripts/cache` directory, like the current `social_media.py` script does for YouTube API credentials. This way, the path is relative and isn't specific to how the person running it manages their computer.
* It'd be nice to update the social media section of the README with a couple lines to mention the new field and the new script.

I think each of these are pretty small tweaks. Game to make them, @dannguyen?

JoshData:
I like IDs!

dannguyen:
@konklone Thanks for the lengthy writeup...I can do those tweaks. Sorry, I'm still a Python noob and can't figure out how to set up my project space properly

konklone:
No problem at all. And if any of them are unexpectedly annoying, let me know and I can address them during the merge process.

dannguyen:
Ah OK, I'm about done but now realize I should just add my changes into `social_media.py` as a new `resolvetw()` method?

konklone:
If it's easy enough, that'd be ideal. If it's not, let me know and we can do it post-merge or just document the new command.

dannguyen:
OK, I've added to social_media.py. Running: 
    
       python social_media.py --resolvetw

Will:

1. Resolve `twitter` entries to Twitter IDs
2. Verify that existing `twitter_id` entries have the correct screen_name as the `twitter` value. If not, the `twitter` value is updated.

I also added [scripts/social/twitter.py] to keep things from getting too cluttered in social_media.py.

konklone:
I merged, after adding two commits:

* https://github.com/unitedstates/congress-legislators/commit/9067af0790f9763be9a977b11d4727ff2e560c28 - I added `tweepy` to `requirements.txt`.
* https://github.com/unitedstates/congress-legislators/commit/c8c972d2953e16fd0de066ccc14b92f2f8a7687e - I noticed your commit had removed a couple of no-longer-valid screen names. I went and replaced them by looking at the members' sites (they'd renamed the accounts), and then ran your new `--resolvetw` command to resolve their IDs. 

The ID resolution worked like a charm, after I created a Twitter app and access token and put them in `scripts/cache/twitter_client_id` as your script's comments told me to.

Thanks for following up and making those changes, @dannguyen! I didn't realize the extent of how helpful the new `--resolvetw` command is -- using the IDs to detect renames and missing/suspended accounts is fantastic, and will make our data even more reliable going forward. :100:

dannguyen:
My pleasure...I use this dataset so much it's the least I can do

PR #312, opened 2015-08-28
Update contact_form urls and spelling correction for Linda Sanchez

davek09:


legislators-current.yaml
```diff
@@ -27274,7 +27274,7 @@
     address: 322 Hart Senate Office Building Washington DC 20510
     phone: 202-224-6542
     fax: 202-228-3027
-    contact_form: http://www.schumer.senate.gov/Contact/contact_chuck.cfm
+    contact_form: http://www.schumer.senate.gov/Contact/email_chuck
```
konklone: Almost - http://www.schumer.senate.gov/Contact/email_chuck takes me to a blank page, but http://www.schumer.senate.gov/Contact/email-chuck works (dash instead of underscore).

konklone:
Looks good - mind making the change I noted, then commenting here to note that it's done? Happy to merge after that.

legislators-current.yaml
davek09: Ok - sorry about that! Will make the change right now.

davek09:
Ok, all done! Thanks!

joelcollinsdc:
Seems like these contact forms urls could be validated against what it says in the contact-congress repo, right?

konklone:
@joelcollinsdc I haven't been involved with this part of it, but I think in theory we do:

https://github.com/unitedstates/congress-legislators/blob/master/scripts/contact_forms.py

Like all of the scripts in this repo, they're applied in an ad hoc fashion. We don't yet have any automation structure that runs them at regular intervals and submits PRs. Instead we rely on production integrators like @JoshData and others to do so at a strong enough rate to keep things high enough quality (with a big burst at session change). Automation would be an upgrade.

konklone:
Looks like the Travis build failure is a Travis fluke:

```
3.4 is not installed; attempting download
/Users/travis/build.sh: line 229: lsb_release: command not found
tar: Unrecognized archive format
tar: Error exit delayed from previous errors.
```

Looks good to me, merging in!

davek09:
Thanks for the merge! I'll take a look at the scripts later to see if there's a nice way to automate this process.

PR #318, opened 2015-10-21
added thomas id for Darin LaHood

dmc2015:


crdunwel:
Looks goods  :+1: . Congress scraper shows correct ID from source http://thomas.loc.gov/cgi-bin/bdquery/?&Db=d114&querybd=@FIELD(FLD004+@4((@1(Rep+LaHood++Darin))+02295))

PR #322, opened 2015-10-27
Adding missing Wikipedia links; adding some Wikidata links

harej:
Adding missing links to Wikipedia for those entries missing Wikipedia
links.

Also, on a trial basis, including links to Wikidata as well. Wikidata
IDs are stable identifiers, meaning they will continue referring to the
member of Congress in question even if the article is renamed. The
compromise reached by myself and Konklone is to list both side by side
for the time being, with the Wikidata ID being used to occasionally
check to see that the Wikipedia page titles are up to date. I will add
Wikidata IDs for the rest of them as soon as I find an efficient way to
do so.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -37267,6 +37267,8 @@
     fec:
     - H4FL19074
     maplight: 2060
+    wikidata: Q16728087
+    wikipedia: Curt Clawson
   name:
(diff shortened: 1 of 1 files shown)


konklone:
Thank you, @harej!

crdunwel:
:+1:

PR #325, opened 2015-10-28
Wikidata IDs and updated Wikipedia article titles for all

harej:
Got Wikidata IDs for the rest of them by starting with the Wikipedia
titles, auditing them to make sure they were referring to the right
person, and then adding the ID in. I updated page titles for articles
that were renamed; in the future, a script will be able to do this
automatically based on the Wikidata ID. The question is to how to do
this automatically in the long term.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -15,6 +15,7 @@
     maplight: 168
     washington_post: gIQA3O2w9O
     icpsr: 29389
+    wikidata: Q381880
   name:
     first: Sherrod
     last: Brown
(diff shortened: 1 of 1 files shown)


konklone:
:+1: Thanks!

Yeah, sometimes it's inevitable that there's an audit phase, or that you need to use two fields to help keep each other up to date. We do that with the social media data. My suggestion is to keep the work to one script, and to provide two different modes. Ideally, it'd be fully automatable, but if not, then something that isn't too onerous to run a few times a year.

PR #331, opened 2015-11-12
highlighted all .yaml files

mohataher:
Some .yaml files were not highlighted which made reading the text a bit difficult. I highlighted them as in-text-code to make it easier reading README.md file.


konklone:
:+1: Thanks!

PR #335, opened 2015-12-17
update donald norcross & bonnie watson coleman

stvnrlly:


stvnrlly:
Rep. Coleman Watson appears to have deleted her Twitter and Instagram accounts. She does have a Flickr account, which it doesn't look like this project tracks.

konklone:
Very strange about Watson Coleman, especially since her official site still links to it. But definitely worth removing, at least.

PR #336, opened 2015-12-19
Valid yaml for Speaker Ryan's twitter

emma-foster:
Placed a space after Speaker Ryan's twitter account name, forming valid yaml.


JoshData:
Thanks!

PR #341, opened 2016-02-15
Remove unused import

hugovk:
This was meant to be in PR #340 but I forgot to push the commit, so here it is!

If you wanted to catch these sooner, how about moving pyflakes from `after_success:` to the end of `script:` in .travis.yml?


konklone:
That sounds good to me!

hugovk:
See PR https://github.com/unitedstates/congress-legislators/pull/342 for that.

PR #342, opened 2016-02-15
catch pyflakes sooner

hugovk:
Helps avoid things like https://github.com/unitedstates/congress-legislators/pull/341 by failing PRs on Travis CI if there are pyflakes problems.

It does this after the main tests, so it you'll always get the results of those first.


konklone:
Looks good, and [the build passed](https://travis-ci.org/unitedstates/congress-legislators/builds/109358041). :+1:

Issue #343, opened 2016-02-18
Adding district office data


pmkane:
Hi all:

We're interested in adding district office data (in-state office addresses/phones/etc for Representatives and Senators) to the dataset.  Reviewing the existing YAML, I can't tell if the preferred approach is to add this data to a separate file (ala legislators-social-media.yaml) or to add it directly to legislators-current.yaml.

Thoughts?

dwillis:
I think district office data would go in a separate file, but curious what others think.

konklone:
I think it could work either way, though I am curious how much the data depends on the particular Rep or Senator in office. Do incoming members just take over the office space of the previous ones, or do they tend to close offices and open new ones?

pmkane:
They open new offices, typically.  They are all leased directly by the Senator or Representative.

Since it sounds like there aren't strong feelings on it either way, we'll submit a pull request as a separate file.

Thanks.

dwillis:
Well, redistricting can definitely change the office locations, and many senators move offices around upon being elected, so things can and do change. I don't think they change a lot, though. The reason I suggested a separate file is that I'm not sure that this info really needs to be in legislators-current.

konklone:
> They open new offices, typically. They are all leased directly by the Senator or Representative.

Congress is weird.

> Since it sounds like there aren't strong feelings on it either way, we'll submit a pull request as a separate file.

:+1:

joelcollinsdc:
@tcarobruce Incredible work.  Are there scripts that you used to gather the data?  How do you envision this is kept up to date?

pmkane:
Hi Joel:

This is a district office database that was maintained by Congress Merge (www.congressmerge.com).  Michael Waters, the owner of Congress Merge, gave us permission to reformat it for congress-legislators and release it under CC license.

Keeping it up to date is an unsolved problem.

JoshData:
All files in this project must be in the public domain. See https://github.com/unitedstates/congress-legislators/blob/master/CONTRIBUTING.md. CC licenses aren't acceptable, unfortunately.

I'm also :-1: on any data that has no update strategy.

pmkane:
Hi Joshua:

Getting it released public domain instead isn't an issue.

But if it won't be accepted without an update strategy, then it's not worth
the round trip to the original vendor for the sign off. Hopefully adding
the data provided impetus, to either us or someone else, to track updates.

Let us know!

[...]

PR #344, opened 2016-03-04
Add district office data

tcarobruce:
This adds district office data for current Members of Congress, as discussed in #343 .

Contains a list of offices for each MOC, with address, phone, and fax (where available). Entries are keyed by bioguide, thomas, and govtrack ids (as is done with the social media data).


tcarobruce:
Sorry, did not intend for the last change to be included in this pull request. I closed it and will open a new request if/when we're able to address the issues in #343.

Issue #276, opened 2015-04-29
New House member data XML file


JoshData:
The House Clerk is publishing a new member data file in XML linked from http://clerk.house.gov/:

http://clerk.house.gov/xml/lists/MemberData.xml

We should replace any scraper with this data where possible.

H/t and big thanks to @GullicksonK!

schmod:
It's a Christmas Miracle!

joelcollinsdc:
I started working on this here:

https://github.com/joelcollinsdc/congress-legislators/commit/16ed6dbc3b69a78420f47424e00e3f2f0050afa2

Some questions:
* Should all data in the Clerk's XML feed take precedence over data in congress-legislators?  Lots of middle names and such are different.  Also lots of first names in the clerk XML feed appear to just be initials?  I am presuming this was their official preference...
* Should I remove the , before Jr. in the offical name to match what is currently in this repo?

JoshData:
I totally forgot about this. Neat.

> Should all data in the Clerk's XML feed take precedence over data in congress-legislators? Lots of middle names and such are different. Also lots of first names in the clerk XML feed appear to just be initials? I am presuming this was their official preference...

I'm not sure. It depends on how everyone uses the data to render names. G. K. Butterworth, for instance, is always trouble. I'll need to look more carefully.

> Should I remove the , before Jr. in the offical name to match what is currently in this repo?

You mean in `official_full`? I added `official_full` only for the purposes of sanity checking that the members were aligning properly against the old unstructured data source. Since this new data has a bioguide ID in it, we should just drop the `official_full` field from our YAML (at least as it exists now --- if we want to put in a full name field, we should rethink what the best content should be).

joelcollinsdc:
These are the name related fields that appear to be available

```
<lastname>Young</lastname>
<firstname>Don</firstname>
<middlename/>
<sort-name>YOUNG,DON</sort-name>
<suffix/>
<courtesy>Mr.</courtesy>
<official-name>Don Young</official-name>
<formal-name>Mr. Young of Alaska</formal-name>
```

There is one member (IL18) that has no official-name for some reason (workflow?).  I imagine there is a lot of thought that goes into this stuff.  It would make sense to me to make these fields available via this repo for those that choose to use them.

GullicksonK:
Joel, 

[...]

PR #345, opened 2016-03-05
Move to import house contacts from the Clerk XML feed

joelcollinsdc:
From #276, the Clerk of the House provides XML for member information.  There are several name fields in the XML that the Clerk provides that do not directly match the name fields in legislators-current.

### DIFF
--- a/scripts/house_contacts.py
+++ b/scripts/house_contacts.py
@@ -6,15 +6,11 @@ import requests
 import lxml
 import re
 from datetime import datetime
-import utils
-from utils import download, load_data, save_data, parse_date
+
+from utils import load_data, save_data, parse_date
 
 def run():
 	today = datetime.now().date()
-
-	# default to not caching
-	cache = utils.flags().get('cache', False)
-	force = not cache
 
 	y = load_data("legislators-current.yaml")
 
--- a/scripts/house_contacts.py
+++ b/scripts/house_contacts.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Update current congressmember's contact info from clerk XML feed
+# Update current House Member's contact info from clerk XML feed
 
 import requests
 import lxml
@@ -22,6 +22,7 @@ def run():
 	xml = requests.get("http://clerk.house.gov/xml/lists/MemberData.xml")
 	root=lxml.etree.fromstring(xml.content)
 
+	yaml_moc_ids = []
 	for moc in y:
 		try:
 			term = moc["terms"][-1]
@@ -34,6 +35,8 @@ def run():
 		if today < parse_date(term["start"]) or today > parse_date(term["end"]):
 			print("Member's last listed term is not current", moc, term["start"])
 			continue
+
+		yaml_moc_ids.append(moc['id']['bioguide'])
 
 		if "class" in term: del term["class"]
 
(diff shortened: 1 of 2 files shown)


scripts/house_contacts.py
```diff
@@ -1,22 +1,24 @@
...
 	today = datetime.now().date()
 
-	# default to not caching
-	cache = utils.flags().get('cache', False)
-	force = not cache
-
 	y = load_data("legislators-current.yaml")
 
+	# TODO use download util?
+	xml = requests.get("http://clerk.house.gov/xml/lists/MemberData.xml")
```
konklone: Could really use an HTTPS URL for this...

konklone:
@joelcollinsdc Mind if we give you write permissions for the repo? You're definitely at the point where we'd welcome you submitting PRs that go branch-to-branch instead of fork-to-branch. It makes it easier for us to hop in and tweak stuff before merging, too.

konklone:
To document for the PR (since the changes are too big to show on the `Files changed` tab), this adds a `list` and `formal` field for each member, like so:

```diff
@@ -6163,6 +6219,8 @@
     middle: R.
     last: Carter
     official_full: John R. Carter
+    list: Carter, John
+    formal: Mr. Carter of Texas
```

Some of the `formal` names have `of [state]` at the end, some don't. @joelcollinsdc, what determines whether they do or not?

@joelcollinsdc: did this commit changing Paul Ryan's name come from the script, or from a manual override?

```diff
@@ -26516,7 +26516,8 @@
   name:
     first: Paul
     last: Ryan
-    official_full: Paul Ryan
+    official_full: Paul D. Ryan
+    middle: D.
```

scripts/house_contacts.py
aboodman: heh

joelcollinsdc:
Sure!   I'll keep any stuff out of master unless told otherwise.

konklone:
Yeah, in general PRs are always preferred for any changes to the repo, to keep visibility high. I sent you an invite to the organization.

scripts/house_contacts.py
joelcollinsdc: no HTTPS version of the clerk URL works at the moment unfortunately.  I'll look into that though.

joelcollinsdc:
@konklone I can check with the Clerk's office on this, from what I understand the Member fills out their preference for their names.  The "D." in "Paul D. Ryan" was an official request, and came from the source XML, no hand editing was done.

konklone:
:+1: Looks good to me.

JoshData:
Hang on there. I un-did the merge. This needs a little more work...

* This doesn't handle nicknames correctly (`middle: C. "Hank"`). Every rep with an existing nickname field needs to be reviewed (e.g. see `Andy Barr`).
* The change in the first name of Scott Rigel (`Edward` => `E.`) is probably going to mess up the name display logic on GovTrack due to his record previously having `Scott` be both his middle name and nickname.
* I'm not sure I'd want to add in the `formal` name. It's House-specific, it's hard to imagine a use case for users of this data, and it's also time-specific (it has e.g. ` of Texas` on names that need disambiguation, which of course only makes sense if you're referring to that person as of the time of the last scrape --- it would be wrong to use that name in conjunction with any historical events, e.g. sponsorship of a bill at a time when the name wasn't ambiguous).
* `list` is basically a duplication of other data fields, it's House-specific, and it isn't in any of the historical records --- so here too I'm not sure I'd want to add this.
* That's mostly also true of `official_full`. I had added `official_full` originally as a sanity check for the current house contacts scraper (because the source data lacked bioguide IDs, iirc). With this PR we don't need that sanity check anymore, so we could consider dropping the field.

joelcollinsdc:
I moved the house-clerk-xml branch into this repo.  Since this PR was closed anyways I'll resubmit and work on it there.

PR #346, opened 2016-03-17
Use House Clerk's new member data XML file instead of scraping contact data

joelcollinsdc:
Work on Issue #276, continuing the discussion in #345

Currently legislators-current has:
first, middle, last, suffix, nickname, official_full

Josh has stated that official_full was only added as a sanity check, which just leaves first and last as the only required fields.

The Clerk's XML has:
- courtesy (ms/mr)
- first
- middle
- last
- suffix
- formal name
- official name
- list name (ex: Nunes, Devin)
- sort name (ex: NUNES,DEVIN)

My understanding of the official versus the first/middle/last fields in the Clerk data is that the official name is the primary data source and the first/middle/last are derived from that, with sometimes confusing results (first names of just a single initial, or middle names that have an initial and then a nickname).  My uninformed opinion would be that just tracking the official name would be the easiest instead of attempting to break it into its various parts.  

I'm not sure I have a preference when it comes to how this gets resolved, just that the less hand-reviewing thats needed would seem to be preferable.


JoshData:
Thanks for continuing the work on the PR. :)

If we're going to have first/middle/last fields in our data, updating it from the House is great. It seems to be working fine except for the nicknames, which is totally fixable. Also the deleting of middle names (which might be good - see below).

I mentioned in the original issue the idea of dropping those fields from our data, but no one else seems to have (see that or) thought it was a good idea.

Alternatively, we could revise the meaning of `middle` and have it be filled in *only* if the Member actually uses it regularly -- i.e. it appears in their official name or on their website. Right now I think we have a lot of middle names for completeness's sake but which presumably no one actually displays. GovTrack's rule is "display the middle name only if the first name is an initial". (What rules do others use?) Making `middle` only present if it's used would simplify the rule (i.e. always use it). Your PR does this anyway/already by deleting `middle` if it's not in the source XML. We could manually clean the Senate, and historical entries might not display as good (by displaying middle names that weren't regularly used) but it would be close enough.

joelcollinsdc:
I thought I would resuscitate this now, I removed all the name matching stuff so the only thing that should be updated are official_full (supposedly only for sanity check anyways?), phone, address, and office

JoshData:
Merged! Thanks for working on this!

PR #352, opened 2016-05-09
Update Sen. Booker official Twitter

RallyCongressStaff:


JoshData:
Thanks! Verified & merging....

PR #353, opened 2016-05-17
Fix details for Cass Ballenger's first term

tmtmtmtm:
Ballenger was elected mid-term (see http://bioguide.congress.gov/scripts/biodisplay.pl?index=B000104, https://en.wikipedia.org/wiki/Cass_Ballenger etc)


JoshData:
Thanks!

PR #357, opened 2016-06-17
added google_entity ids

dwillis:
Google has a Knowledge Graph for its search data that, among other things, has unique entity ids for people, organizations and "things," members of congress among them. I've matched current members to their entity ids (Warren Davidson, the newest member, doesn't have a unique one yet) and added `google_entity_id` to legislators-current.yaml, as well as a script to generate them using the ProPublica API.

These IDs could be used in combination with the Google Trends API to gauge search interest in members, for example. If this seems valuable, many former members also have these and I could work on adding those, too.

### DIFF
--- a/scripts/google_entities.py
+++ b/scripts/google_entities.py
@@ -3,7 +3,6 @@
 # Update current Google entity IDs using ProPublica Congress API.
 
 import json, urllib.request, urllib.parse, urllib.error
-import rtyaml
 from utils import load_data, save_data
 
 def run():
--- a/scripts/google_entities.py
+++ b/scripts/google_entities.py
@@ -8,7 +8,7 @@ from utils import load_data, save_data
 
 def run():
     # load in current members
-    y = load_data("/Users/DW-Admin/code/congress-legislators/legislators-current.yaml")
+    y = load_data("legislators-current.yaml")
     for m in y:
         response = urllib.request.urlopen("http://propublica-congress.elasticbeanstalk.com/represent/api/v1/members/%s.json" % m['id']['bioguide']).read()
         j = json.loads(response.decode("utf8"))
--- a/scripts/google_entities.py
+++ b/scripts/google_entities.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+
+# Update current Google entity IDs using ProPublica Congress API.
+
+import json, urllib.request, urllib.parse, urllib.error
+import rtyaml
+from utils import load_data, save_data
+
+def run():
+    # load in current members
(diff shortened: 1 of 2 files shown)


scripts/google_entities.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Update current Google entity IDs using ProPublica Congress API.
+
+import json, urllib.request, urllib.parse, urllib.error
+import rtyaml
+from utils import load_data, save_data
+
+def run():
+    # load in current members
+    y = load_data("/Users/DW-Admin/code/congress-legislators/legislators-current.yaml")
```
konklone: This should be a relative path.

dwillis: Guh, yes. will fix.

scripts/google_entities.py
```diff
@@ -0,0 +1,21 @@
...
+    for m in y:
+        response = urllib.request.urlopen("http://propublica-congress.elasticbeanstalk.com/represent/api/v1/members/%s.json" % m['id']['bioguide']).read()
```
konklone: Is this endpoint part of the official API? This domain isn't the one officially, documented, which is:  https://api.propublica.org/congress/

That host also has the benefit of HTTPS, which should be a near-requirement for our scripts whenever possible, given the value of data integrity for this project's role in providing upstream source data.

konklone:
I'm a fan of including this. However, how much work does the Pro Publica API do in preparing this data? Would it be feasible to have this script pull from the same source the Pro Publica API does?

scripts/google_entities.py
dwillis: Yeah, I'll do that too.

dwillis:
It's feasible, yes - we hit the Knowledge Graph API to do this, but there's some review/disambiguation involved, since some of the names are common.

konklone:
@JoshData Any opinion?

JoshData:
Lemme look tomorrow.

JoshData:
Neat. +1. Two things:

Warren Davidson has an empty-string id --- let's just remove that line?

And maybe add some example uses to the README?

dwillis:
Sure, can do. I'll re-work the script and clean things up as suggested.

scripts/google_entities.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+
+# Update current Google entity IDs using ProPublica Congress API.
+
+import json, urllib.request, urllib.parse, urllib.error
+import rtyaml
```
konklone: The build is failing because this import statement is unused. I'll fix this on-branch myself.

JoshData:
Superseded by #371.

PR #371, opened 2016-10-20
script to sync wikipedia, wikidata and knowledge graph ids with wikidata, fix bad wikipedia data

thinkcontext:
I noticed @dwillis was adding the Google Knowledge Graph ids to the current legislators in another pull request, so I thought about doing it for the historical ones as well.  I wrote a script to add or update Wikipedia, Wikidata and Knowledge Graph ids joining with Wikidata on the Bioguide id. 

This PR has the script as well as running it against both historical and current legislator files.  If you look at the diff for the legislators you'll find it reveals one actual current mistake, Dan Sullivan's Wikipedia page is linked to a state rep from Arkansas instead of the Senator from Alaska.  Then there are a bunch of places where Wikipedia article names have changed, both in current and historical.  Most of these are just minor formatting, the current ones work fine w/ a redirect but there are some that go to disambiguation pages (eg David Young) so updating them is necessary.


konklone:
:+1: from me - @JoshData ?

JoshData:
Could you just reset the two files in `scripts` to Unix-style line endings?

JoshData:
(Tagging #322 as the source of the error in Dan Sullivan's wikipedia entry.)

JoshData:
Does this cover everything in #357?

thinkcontext:
Fixed the line endings.  Yes, it does cover #357.  I noticed a few places where we differ but spot checking it appears this PR is correct ( @dwillis check out John Carney, Steve Pearce and others).

Also, just to note on the usefulness of Knowledge Graph ids, the Youtube API accepts them.

JoshData:
Fantastic. If there are no objections, I'll merge.

PR #376, opened 2016-11-10
fix bug in wikidata_bioguide.py that caused google_entity id duplicat…

thinkcontext:
…ion. also fix the data the bug messed up.


JoshData:
I still get a dup in H000570/H000571.

JoshData:
Thanks!

PR #390, opened 2016-12-15
Clarifies intent of our social media data

konklone:
Adds an explanation of what our social media data represents to the README.


JoshData:
+1

PR #391, opened 2016-12-17
senate membership updates for the 115th Congress

JoshData:
This is a draft of senate changes for the 115th Congress.

Bioguide IDs for the new senators have not yet been assigned so that's missing. Otherwise it should be good to merge on Jan 3. If something else comes ahead of this PR, the GovTrack IDs might need to be re-done.

I ran:

* a new script scripts/election_results_senate_2016.py
* scripts/sweep.py (renamed from sweep_memberships and now also sweeps the social media file)

see #375 and h/t to @mrumsky who started the senate elections google doc that I started from


konklone:
Approved, but then I noticed tests were failing for what look like legit reasons.

JoshData:
Yeah there's a datetime.now() call in the new validation script that won't let it pass until Jan 3 actually comes around, and the missing bioguide ID.

PR #393, opened 2016-12-21
house membership updates for the 115 congress

joelcollinsdc:
see #375

work in progress... still missing:

* ~~bio~~
* ~~official_full (not in XML yet?)~~ (maybe don't include?)
* ??


scripts/election_results_house_2016.py
```diff
@@ -0,0 +1,119 @@
...
+	district = int(xml_member.find("statedistrict").text[2:])
+	# Add a new term.
+	p['terms'].append(collections.OrderedDict([
+		("type", "rep"),
+		("start", "2017-01-03"),
+		("end", "2019-01-03"),
+		("state", mi.find('state').get('postal-code')),
+		("district", district),
+		("party", party),
+		("caucus", caucus),
```
JoshData: The `caucus` field should only be added if it's different from `party`.

JoshData:
Think this is ready?

Whatever isn't working I plan to just finish up on the 3rd.

joelcollinsdc:
Beyond running the validate code, I didn't do much QA on it, but yes I
think its readyish.  Would it make sense to not merge this and instead wait
on the official House XML file (on the 3rd?), re-run it, and then do the
finishing up on the 3rd?

JoshData:
Yeah, if the official XML is posted promptly, that makes sense.

JoshData:
Actually, because I want to get this done quickly, I'm going to merge this based on the data we have today. If there are changes in the official XML, we can update the data at that point.

I'm going to force-push an update to this branch that rebases it on my Senate PR so that it's ready to go tomorrow.

jlev:
@JoshData thanks for the quick action on the merge. I use legislators-current.yaml for a call-congress tool, so it's great to have updated info right at the start of the term.

However, I'm noticing that some of the new terms are missing the phone field. This is the case for 34 Senators, a list is attached. Looking through election_results_senate_2016.py, I don't see a line to copy forward phone numbers for senate incumbents. Should 'phone' be added to the list on line 69?

[115-terms-missing-phone.txt](https://github.com/unitedstates/congress-legislators/files/682926/115-terms-missing-phone.txt)

jlev:
I went through and pulled forward existing phone numbers from incumbent senators previous terms. For new offices, I called the number listed at http://www.senate.gov/general/contact_information for the same seat, and both Hassan and Harris are connecting through to the new staff. No one picked up at Vitter's old number, but maybe Kennedy's staff are slow getting in today.

Here's an update for the yaml file: https://github.com/spacedogXYZ/call-power/commit/8c5245c8e8d7773b1edc2d06a0273443c3c27ca7#diff-63ee354536477331df94df3c35d6b018

I can also go through and adjust the generation scripts, but I also don't want to make changes if someone else is responsible for the updates.

Thanks again for making high quality open data available in a post-Sunlight world.

JoshData:
We have a script to do the update from official data. I ran the script & updated the files.

Issue #387, opened 2016-12-13
Add death dates


bycoffe:
One reason for some of the incorrect term end dates for legislators in the historical file is death in office, when a member died but the term end date is listed as the end of the full term. Adding the date of death (under the `bio` key) would allow us to compare each person's death date and last term end date to find those that should be fixed.

I'd propose using Wikidata as the source for the dates.

LeafyLi:
Hello, I'm a beginner, how can I help with this or any other part of the project?

GPHemsley:
Has any work been done on this?

(Given that the existing key in `bio` is `birthday`, what would the death date key be?)

CopOnTheRun:
Hi I think I should be able to help out on this front if the project's maintainers are amenable to it. I scraped the wikipedia page of every congressperson before I knew that this project existed, and death was one of the fields I grabbed. It would be easy enough for me to update the information in the yaml. I'd also be willing to make a separate commit correcting all the terms that end after a person's death.

The word "deathday" isn't super common, but I think it's self evident enough when compared to "birthday".

JoshData:
I'd first be interested in correcting term end dates. That would be really helpful to improve the accuracy of the data here.

Death dates are interesting but would be hard to maintain going forward.

PR #394, opened 2016-12-27
Working on #387.  Adding entries from https://en.wikipedia.org/wiki/L…

eliotpearson:
Pulled the latest year of deceased congressman and added them based off of:

https://en.wikipedia.org/wiki/List_of_members_of_the_United_States_House_of_Representatives_who_died_in_2016

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ The following fields are available in `legislators-current.yaml` and `legislator
 	* birthday: The legislator's birthday, in YYYY-MM-DD format.
 	* gender: The legislator's gender, either "M" or "F". (In historical data, we've worked backwards from [history.house.gov's Women in Congress feature](http://history.house.gov/People/Search?filter=6).)
 	* religion: The legislator's religion.
+	* died: The legislator's date of death if applicable.
 
 * terms (one entry for each election)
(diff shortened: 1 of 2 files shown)


JoshData:
Cool. I think `died` would be a better key name. Also need to document the new field in the README.

eliotpearson:
@JoshData thanks for the quick PR review.  How about this?

JoshData:
So now the question is, how do we get this info for all of the deaths Wikipedia has?

House: 1970 1971 1972 1973 1974 1975 1976 1977 1978 1979 1980 1981 1982 1983 1984 1985 1986 1987 1988 1989 1990 1991 1992 1993 1994 1995 1996 1997 1998 1999 2000 2001 2002 2003 2004 2005 2006 2007 2008 2009 2010 2011 2012 2013 2014 2015 2016
Senate: 1910s 1920s 1930s 1940s 1950s 1960s 1970s 1980s 1990s 2000s 2010s

A scrape of bioguide.congress.gov might be more effective, since we could get every death date in one shot. See https://github.com/unitedstates/congress-legislators/blob/bioguide-deep-parse/scripts/bioguide.py for some work I did on that. It might already be able to do it.

eliotpearson:
Unfortunately I do not have bandwidth to do this.  Learned a lot from this project.  Good luck.

PR #395, opened 2017-01-03
Rep. Candice Miller resigned effective Dec. 31, 2016

bycoffe:
Source: http://clerk.house.gov/member_info/vacancies.aspx


JoshData:
Wat. Why would she resign three days before the term ends?

bycoffe:
She was elected public works commissioner (http://www.freep.com/story/news/local/michigan/macomb/2016/11/08/macomb-elections-public-works-marrocco-miller-hackel/93444650/); I'm assuming her term for that started Jan. 1 and she had to resign from Congress first.

This can be folded into the 115th Congress update if that's easier; just want to make sure her end date is correct.

PR #397, opened 2017-01-06
Votesmart ids, wikidata, wikipedia for new congress

thinkcontext:
Updated wikidata_bioguide.py to get Votesmart and Opensecrets ids from Wikidata, then ran it against the latest legislators-current.  It picked up Votesmart for new congress plus a bunch of existing ones as well.  It also picked up Wikidata and Wikipedia for the new congress, plus a few google ids.

It looks like Opensecrets hasn't created ids for the new Congress yet.  We can rerun this as they make it into Wikidata or we can write a different script to import it from their API.


JoshData:
Could you find / let me know a way to spot check the wikidata and google entity IDs? Where would I plug those IDs into?

dwillis:
For Google IDs, you can test them here, but you'll need to drop the kg:/ bit (we should probably do this for all of them, as it doesn't appear necessary): https://www.google.com/trends/explore

thinkcontext:
Sure.

https://www.wikidata.org/wiki/Q10853588
https://g.co/kg/m/08sry2

For the google id note that the colon is removed.

JoshData:
Thanks. I was going to spot check a few but I don't have time so just merging now.

PR #399, opened 2017-01-14
Update House websites, a Senate website, and some data source URLs

konklone:
I ran `house_contacts.py` and `senate_websites.py`, and updated them (and the `committee_membership.py` script) to use `https://` URLs for `house.gov` and `senate.gov`.


JoshData:
+1

konklone:
@JoshData Should I take that as an approved review? :)

PR #403, opened 2017-01-24
Peter Welch IG peterwelchvt => reppeterwelch

msimonborg:
fix inaccurate instagram handle for Rep. Peter Welch of VT


joelcollinsdc:
sorry @msimonborg I didn't see this even though it was first, #404 was merged already.  thanks!

PR #413, opened 2017-01-30
URL and contact form for Sen Kamala Harris - CA

msimonborg:
hopefully this works, and doesn't throw in all the extras you don't need at the moment


joelcollinsdc:
thank you, since there was a conflict I took the liberty to create a new PR #419  with this change and some others.  thanks.

PR #419, opened 2017-02-01
update senate contact forms

joelcollinsdc:
There seemed to be many issues with contact forms, this hopefully will fix most of them, but I question the value of keeping this list updated here as opposed to pointing people to https://github.com/unitedstates/contact-congress

The script got most of them, but the senate XML will often point to a contact landing page instead of the form itself, so its not totally reliable and this is a manual review process.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -416,7 +416,7 @@
     address: 393 Russell Senate Office Building Washington DC 20510
     phone: 202-224-6324
     fax: 202-228-0604
-    contact_form: http://www.casey.senate.gov/contact/
+    contact_form: https://www.casey.senate.gov/contact/
     office: 393 Russell Senate Office Building
     state_rank: senior
     rss_url: http://www.casey.senate.gov/rss/feeds/?all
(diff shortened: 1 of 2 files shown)


JoshData:
I'm OK with moving the field to the other repo (I don't use the field). Also +1 on this PR (since I don't use the field <g>).

joelcollinsdc:
Could someone comment then on what this field is used for?  Is blindly copying the <email> field from the senate XML acceptable?  The difference is... sometimes this field points to a contact landing page instead of an email form.  Depending on the use case this may be more preferable anyways.

konklone:
I'm also :+1: on removing it and seeing if anyone complains, though yes, this does serve a different function than contact-congress and it is conceivable that people would use it for different purposes. 

@joelcollinsdc Did you manually review URLs and change landing pages to the contact forms, when the Senate XML gave you a landing page? I would say that, if that's what's required to make these useful, it is not worthwhile. I would be more okay with linking users to landing pages and increasing their clicks, than putting a manual review step on the maintainers of this field.

joelcollinsdc:
I changed the logic to just copy whatever it says in the XML.  i did notice one error where the page it says in the XML is a 404 so I left that...

-    contact_form: http://www.klobuchar.senate.gov/public/email-amy
+    contact_form: http://www.klobuchar.senate.gov/public/contact-amy

I like this approach much better.

konklone:
Awesome, I added a tiny typo commit, and it looks good to me. Merging.

PR #421, opened 2017-02-06
Updated Twitter and Facebook social media accounts

konklone:
This does a full sweep of Twitter and Facebook accounts. This scans MoC house.gov and senate.gov pages (for any MoC where we don't have an account for them already) and looks for links to twitter.com and facebook.com links, subject to filtering out common false positives.

I went and hand-verified that every Twitter account either links back to the house.gov/senate.gov site (or was otherwise clearly not a campaign account and positively linked to from the house.gov/senate.gov site), and hand-verified that every discovered Facebook account either has a verified checkmark for a Page set to Government Official, or that it links back to the homepage or is otherwise clearly not a campaign account. I also sanity checked every Twitter and Facebook handle to make sure it matched up with the name of the member in question.

I did remove one -- [Congressman Soto](https://soto.house.gov/)'s [Facebook account](https://www.facebook.com/people/US-Rep-Darren-Soto/100014901039474) -- which doesn't link properly through our scanner, as it's set to a personal account and not a Page like everyone else. I'll give him some time to change that before attempting to fix or otherwise blacklist the page.

### DIFF
--- a/legislators-social-media.yaml
+++ b/legislators-social-media.yaml
@@ -1207,6 +1207,7 @@
     youtube: RepDavidPrice
     youtube_id: UCtX7bSXBLF9ge-Q955n5Vqw
     twitter_id: 155669457
+    facebook: RepDavidEPrice
 - id:
     bioguide: P000449
     thomas: '00924'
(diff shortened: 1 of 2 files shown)


joelcollinsdc:
LGTM, which script is this, social_media.py?

konklone:
Yep. `./social_media.py --sweep --service=twitter`, then go and validate the contents of the generated CSV in `cache/`, then `./social_media.py --update --service=twitter`. Then same with `facebook`.

PR #424, opened 2017-02-09
Replace Jeff Sessions with Luther Strange

bycoffe:
This moves Jeff Sessions, who resigned, to `legislators-historical`. It adds Luther Strange, who was appointed to replace him, to `legislators-current`. Not sure if the `end_date` for Strange's term should be handled differently since it's a temporary appointment until there's a special election.


JoshData:
For consistency Strange should be added to the end of the file. Otherwise looks good.

If you wouldn't mind assigning the next available GovTrack id `412734` that'd save me a step.

bycoffe:
@JoshData done, thanks

PR #426, opened 2017-02-10
Fixed the terms for TX Sens Gramm and Cornyn

patrickvankessel:
Gramm's resignation was effective 11/30/2002, and Cornyn served in the interim - 11/30/2002 to the end of the session, 01/03/2003


JoshData:
Nice!

Issue #411, opened 2017-01-29
Drop fax #s?


JoshData:
In #219 I suggested we drop fax numbers, but consensus was that we should keep them and Sunlight suggested they were going to keep them up to date.

Per #405, they may be out of date. I don't know if anyone is checking them. Sunlight obviously isn't anymore. Our scripts don't update the values. The only fax numbers we're now offering is for senators who were serving last Congress - we didn't pull forward fax numbers for re-elected members.

If no one chimes in to say they are taking responsibility for the fax numbers, I'd like to drop them.

essandess:
With respect, absolutely not. 👎🏻👎🏻👎🏻

For effectiveness, the fax data is more important than the email forms.

The advice I read about contacting Congress is:

1. Phoning a district office is most effective, https://storify.com/editoremilye/i-worked-for-congress-for-six-year
2. Writing a district office is most effective, https://www.ted.com/talks/omar_ahmad_political_change_with_pen_and_paper
3. Writing or phoning the personal contact of the LA is most effective, like the lobbyists all do
4. Faxing a personalized letter is as good as a mail letter
5. Email the least effective, https://storify.com/editoremilye/i-worked-for-congress-for-six-year
6. Social media not effective at all

Personally, I choose to go with fax because it's as fast as an email and more effective, fax modems cost under $15 https://www.amazon.com/Internet-Modem-V-90-V-92-Sewell/, and my calling plan already covers all US long distance calls.

If I had more time I'd call the district offices. But I don't, so I fax rather than doing nothing.

It would be great if the EFF's https://democracy.io [EFForg/democracy.io](../../../github.com/EFForg/democracy.io) or Sunlight's [contact-congress](../../../unitedstates/contact-congress) offered the range of phone/fax/email contact options.

essandess:
I haven't checked all the fax numbers in the existing data, but the ones I have (numbers in the hundreds) have all been correct, so this looks like a useful, accurate dataset to me.

essandess:
And per #405, these are all ***additions*** of fax data to the district offices, not corrections, so no, they are not out of date.

JoshData:
I strongly doubt that a faxed letter is better than a message sent through a webform (which is what I assume you mean by an email).

But that aside - If you're volunteering to do the work, then keeping the fax numbers is good with me. Will you submit a PR that ensures every legislator has an up to date fax number listed?

JoshData:
[...]

PR #428, opened 2017-02-11
Remove spaces from fax numbers, unique lookups, current term filter

essandess:
Minor updates to database and lookup code

### DIFF
--- a/congress_lookup.py
+++ b/congress_lookup.py
@@ -11,7 +11,7 @@ __author__ = 'stsmith'
 
 # Author 2017 Steven T. Smith <steve dot t dot smith at gmail dot com>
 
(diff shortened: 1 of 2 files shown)


konklone:
Thank you for doing this!

@essandess Would you be up for also modifying the script that produces this data to also automatically remove spaces from numbers, so that the issue doesn't reappear in future?

essandess:
@konklone I'll take a look at the script, but I'm not familiar with your stack. Would you please point me to it?

konklone:
My bad, I mis-remembered this being initialized by script, but we've managed them in batches (see #236 and #238) because they are so slow-changing (as you noted in https://github.com/unitedstates/congress-legislators/issues/411#issuecomment-279192285).

:+1: from me, merging!

PR #430, opened 2017-02-12
pull fax numbers from 114th forward

joelcollinsdc:
See #411


konklone:
Thanks @joelcollinsdc and @essandess!

PR #431, opened 2017-02-13
fax additions; sorted output, filter method, keys existence test

essandess:
Add a few new fax numbers, fix some fail lookup conditions and add a filter method for possible future use.

### DIFF
--- a/congress_lookup.py
+++ b/congress_lookup.py
@@ -61,8 +61,8 @@ class CongressLookup:
     def lookup_by_member(self,property,member):
         for leg in ( leg for leg in self.legislators if \
                     (leg['name']['official_full'] == member['name']) \
-                    or ('bioguide' in leg['id'] and leg['id']['bioguide'] == member['bioguide']) \
-                    or ('thomas' in leg['id'] and leg['id']['thomas'] == member['thomas']) ):
+                    or ('bioguide' in leg['id'] and 'bioguide' in member and leg['id']['bioguide'] == member['bioguide']) \
+                    or ('thomas' in leg['id'] and 'thomas' in member and leg['id']['thomas'] == member['thomas']) ):
             self.lookup_legislator_properties(property,leg)
 
     def lookup_by_lastname(self,property):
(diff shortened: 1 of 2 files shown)


congress_lookup.py
```diff
@@ -61,8 +61,8 @@ def inclusive_wildcard_match(self,name,pat):
     def lookup_by_member(self,property,member):
         for leg in ( leg for leg in self.legislators if \
                     (leg['name']['official_full'] == member['name']) \
-                    or ('bioguide' in leg['id'] and leg['id']['bioguide'] == member['bioguide']) \
-                    or ('thomas' in leg['id'] and leg['id']['thomas'] == member['thomas']) ):
+                    or ('bioguide' in leg['id'] and 'bioguide' in member and leg['id']['bioguide'] == member['bioguide']) \
+                    or ('thomas' in leg['id'] and 'thomas' in member and leg['id']['thomas'] == member['thomas']) ):
```
konklone: So you know, you can do `dict.get('key')` instead of `dict['key']`, and the `.get()` version will just return `None` by default if the key doesn't exist, rather than throwing an exception.

You can also `dict.get('key', 'default value')` and have it return `'default value'` (which can be anything) if the key doesn't exist, rather than `None`.

legislators-current.yaml
```diff
@@ -34606,6 +34608,7 @@
     district: 2
     party: Republican
     phone: 202-225-1621
+    fax: 202-225-2563
```
konklone: This (for some reason) broke the build: https://travis-ci.org/unitedstates/congress-legislators/builds/201021210

```
The command "python test/are_files_linted.py" exited with 1.
```

```
$ python test/are_files_linted.py
legislators-current.yaml needs to be linted:
--- in repository
+++ after linting
@@ -34852,7 +34852,7 @@
     district: 2
     party: Republican
     phone: 202-225-1621
-    fax: 202-225-2563 
+    fax: 202-225-2563
     url: http://wagner.house.gov
     rss_url: http://wagner.house.gov/rss.xml
     address: 435 Cannon HOB; Washington DC 20515-2502
```

I'm not sure why - could there be a tab instead of spaces, or something?

congress_lookup.py
essandess: Yes, but syntax and thinking about how `None` behaves in the subsequent tests...

legislators-current.yaml
essandess: Tricksy trailing space fixed.

FYI if you ever see this again -- I'd suggest that a trailing space shouldn't throw an error.

konklone: > FYI if you ever see this again -- I'd suggest that a trailing space shouldn't throw an error.

It may not matter with YAML, I'm not sure. But I would also suggest using an editor that auto-clears trailing space. :)

PR #434, opened 2017-02-13
social media updates

joelcollinsdc:
updates based on @soooh's comment in #293.

This is probably OK to go after some light review, but social media in general still needs a lot of scrutiny.

### DIFF
--- a/legislators-social-media.yaml
+++ b/legislators-social-media.yaml
@@ -5451,6 +5451,7 @@
     govtrack: 412704
   social:
     facebook: rogermarshallmd
+    twitter: RepMarshall
 - id:
     bioguide: E000296
     govtrack: 412677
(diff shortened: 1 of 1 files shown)


legislators-social-media.yaml
```diff
@@ -5450,3 +5451,46 @@
...
+    bioguide: E000296
+    govtrack: 412677
+  social:
+    twitter: RepDwightEvans
+- id:
+    bioguide: R000609
+    govtrack: 412692
+  social:
+    facebook: RepRutherfordFL
+    twitter: RepRutherfordFL
```
konklone: This is linked to from https://rutherford.house.gov, and it comes up now when I run our script. I guess it was added in the last week. I'll do another sweep and see if other stuff comes up.

konklone: Yeah, another sweep picked up Rep. Rutherford, Dwight Evans, and Mike Johnson. Those are all included here.

PR #440, opened 2017-02-22
add twitter handles for 2 legislators and run resolvetw to fix some twitter ids

joelcollinsdc:
Adds twitter handles for Amash, Hanabusa, Rand Paul, and Rohrabacker, and run social_media.py --resolvetw.  resolvetw showed we had some incorrect twitter ids.

Amash's twitter wasn't linked from his website but the twitter account links to his facebook which links to his official website, which seemed official enough.

### DIFF
--- a/legislators-social-media.yaml
+++ b/legislators-social-media.yaml
@@ -1100,12 +1100,12 @@
     thomas: '01910'
     govtrack: 412308
   social:
-    twitter: JaredPolis
+    twitter: RepJaredPolis
     facebook: jaredpolis
     youtube: JaredPolis31275
     facebook_id: '53481427529'
     youtube_id: UCJJ1Ye0mOzqZh1_GErFUISw
-    twitter_id: 15361570
+    twitter_id: 463132556
 - id:
     bioguide: P000597
     thomas: '01927'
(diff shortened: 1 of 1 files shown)


konklone:
I'm not as confident about the `justinamash` handle. The Twitter account links to his Facebook account, which does link to his official page but which also has the `Politician` category rather than the `Government Official` category ([example](https://www.facebook.com/NancyPelosi/)).

I think it's worth a phone call to his office to ask their staff to confirm the Congressman's official Twitter account, if one exists.

joelcollinsdc:
ugh.  Sorry @konklone   now i see why this repo is based on what's linked on their official website.

Rep Paul's situation is the same, it pretty clearly appears to be the twitter account the office is actively using, but its also the account used for campaigning.   So I removed Paul and Amash.  :(

The other 2 accounts, Rohrabacher says official on the twitter account, and hanabusa has a separate campaign account even though housedemocrats is still linked from the website so I think those are OK.

Accounts with a ton of followers are probably less likely to create a new 'official' account b/c they would lose all those followers.  And since the president just did it, (i know, different rules apply), is it going to be more common?  Seems like we are missing a pretty big piece of congressional chatter by leaving these out.  

Also FWIW, the cspan members-of-congress twitter list contains these campaign accounts...

joelcollinsdc:
I ran verify and 4 accounts showed up as being different than the website.  Taking off CoryBooker and replacing with senbookeroffice was particularly... unfortunate

konklone:
Thanks for doing that. I'm disappointed I missed the commit in #352 where CoryBooker's handle was introduced. I'm keeping a closer eye on them now than I was then -- thanks for your help, @joelcollinsdc.

PR #441, opened 2017-02-24
prefer existing social media over first found

joelcollinsdc:
I noticed when running social_media.py --verify [--resolvetw], candidate_for is currently returning the first regex that passes.  This changes the logic to return the existing candidate first if its found.  This causes the output of the script to be less verbose, like so:

[W000437] mismatch on http://www.wicker.senate.gov - SenatorWicker -> AjitPaiFCC

Also, would anyone object to making lines like this debug-only?
Found redirect, downloading http://www.thune.senate.gov/public/ instead..

### DIFF
--- a/scripts/social_media.py
+++ b/scripts/social_media.py
@@ -433,7 +433,7 @@ def main():
 
       bioguide = entry['id']['bioguide']
 
-      candidate = candidate_for(bioguide)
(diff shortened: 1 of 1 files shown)


konklone:
:+1:, and sure, making that line debug only sounds just fine.

PR #442, opened 2017-02-27
facebook and instagram sweeps, senate contacts

joelcollinsdc:
Ran senate contacts and facebook/ig sweeps.  

 did not run resolveig or resolvefb... I spent a few minutes trying to get the APIs working and gave up... if anyone has instructions on how to get these APIs working it would be good to get documented.  

For the instagram accounts, I checked accounts not starting with Rep or Sen for obvious campaign accounts, only found one.  I did not check them if they sounded official.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -359,7 +359,7 @@
     state: DE
     party: Democrat
     class: 1
-    url: http://www.carper.senate.gov/public
+    url: https://www.carper.senate.gov/public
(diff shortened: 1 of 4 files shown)


konklone:
The `resolveig` and `resolvefb` tasks probably need updating for IG's and FB's current APIs. They've always been more brittle than the rest of the task (surprisingly, since they're API-based and not about scraping...).

In any case, this looks good. :+1:

JoshData:
Nice.

PR #445, opened 2017-02-28
run wikidata_bioguide.py

thinkcontext:
Added getting ballotpedia from wikidata to the script and ran it.  Picked up several ballotpedia ids, updated a few renamed wikipedia articles and added a mess of google entity ids.


konklone:
Well, I can't see why the Travis build failed, perhaps because of the S3 outage or a second-order outage, but would you mind fixing the build before we merge? You can replicate it by running the commands in `.travis.yml`.

joelcollinsdc:
Restarted the build and it passed.

PR #436, opened 2017-02-15
Update House committee membership using House XML data

joelcollinsdc:
Probably some review needed here, especially around subcommittees and joint committee handling.  Refs #276

```
committee_membership_sanity_check.py
HLIG    : 22
 HLIG01 : 10
 HLIG02 : 10
 HLIG03 : 10
 HLIG04 : 10
HSAG    : 46
HSAP    : 52
HSAS    : 62
HSBA    : 60
HSBU    : 33
HSED    : 39
HSFA    : 47
HSGO    : 38
 HSGO06 : 9
 HSGO24 : 14
 HSGO25 : 8
 HSGO27 : 10
 HSGO28 : 6
 HSGO29 : 8
HSHA    : 9
HSHM    : 30
HSIF    : 55
 HSIF02 : 17
 HSIF03 : 31
 HSIF14 : 31
 HSIF16 : 29
 HSIF17 : 22
 HSIF18 : 22
HSII    : 42
HSJU    : 41
 HSJU01 : 13
 HSJU03 : 26
 HSJU05 : 13
 HSJU08 : 17
 HSJU10 : 8
HSPW    : 61
 HSPW02 : 30
 HSPW05 : 37
 HSPW07 : 14
 HSPW12 : 48
 HSPW13 : 13
 HSPW14 : 32
HSRU    : 13
HSSM    : 18
 HSSM23 : 3
 HSSM24 : 4
 HSSM25 : 6
 HSSM26 : 6
 HSSM27 : 5
HSSO    : 9
HSSY    : 31
HSVR    : 21
 HSVR03 : 12
 HSVR08 : 8
 HSVR09 : 6
 HSVR10 : 9
HSWM    : 38
 HSWM01 : 11
 HSWM02 : 17
 HSWM03 : 11
 HSWM04 : 16
 HSWM05 : 15
 HSWM06 : 11
JCSE    : 13
JSEC    : 17
JSLC    : 9
JSPR    : 8
JSTX    : 9
SCNC    : 6
SLET    : 6
SLIA    : 15
SLIN    : 19
SPAG    : 17
SSAF    : 21
 SSAF13 : 0
 SSAF14 : 0
 SSAF15 : 0
 SSAF16 : 0
 SSAF17 : 0
SSAP    : 31
 SSAP01 : 13
 SSAP02 : 19
 SSAP08 : 7
 SSAP14 : 13
 SSAP16 : 18
 SSAP17 : 15
 SSAP18 : 20
 SSAP19 : 17
[...]

PR #448, opened 2017-03-04
WIP: joint house committee membership updates

joelcollinsdc:
Hopefully this is acceptable...  House Joint committee memberships are loaded from the committee-memberships-current.yaml file.  This PR uses a CSV to load this data instead, which I thought made it easier to separate what entries are added manually versus imported from some other location.

This PR also includes removing the thomas id which was discussed in #436 

WIP because I did not add house members for JSLC or JSTX yet because the CHA pages on them still reference the 114th congress...  Not sure if these should just be brought forward or not.

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -10089,43 +10089,6 @@ JCSE:
   rank: 4
   bioguide: S001181
   chamber: senate
-- name: Christopher H. Smith
-  party: majority
-  rank: 1
-  title: Co-Chairman
-  bioguide: S000522
-  thomas: '01071'
-  chamber: house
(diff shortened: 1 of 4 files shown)


JoshData:
All of the files are a mix of hand and automated editing. Not sure there's really a point in splicing off the House members into a CSV file. We can just edit the original by hand? But I don't feel strongly about it.

joelcollinsdc:
Maybe just my personal preference... I think when most of the file is
updated automatically it seems easy for these manual updates to get lost in
the fold.  Also I guess ranks rarely change but its easier to reorder,
switch majority/minority, etc this way.

Issue #362, opened 2016-07-14
yaml vs. json speed


jonathanstrong:
Is there any reason to use yaml for the serialization format? It's super slow. I just installed the C LibYAML bindings and it's still taking my machine ~14 seconds to load legislators-historical.yaml. By contrast if I save the data in json it takes 375ms to open.

dwillis:
Yes, that's the reason. I don't think we envisioned this file as being loaded on the fly by web requests, for example, but rather one that was periodically updated and stored.

jonathanstrong:
makes sense, and yaml is good for that reason. just find it annoying when I go to parse it. thanks!

Hardmath123:
Would a JSON mirror help? The YAML can be serialized easily to JSON, so a git hook that makes JSON copies wouldn't be hard to do. It can live in the gh-pages branch so that people can query the JSON as a web API.

schmod:
The `git blame` is also a lot cleaner with YAML, because commas don't need to be inserted.

AlJohri:
for usage, I just convert to JSON first:

```
yaml2json <(curl -s https://raw.githubusercontent.com/unitedstates/congress-legislators/master/legislators-current.yaml 2>&1) | jsonpp >legislators-current.json
```

requirements
```
npm install -g yamljs
brew install jsonpp
```

hboylan:
I'm using the same yamljs package as @AlJohri.

Unfortunately, it's unable to parse multiline strings like in [committees-current.yaml](https://github.com/unitedstates/congress-legislators/blob/master/committees-current.yaml#L36). 

For now, I have a script to condense multilines into one, but since the YAML should be both human-readable and machine-parseable the correct solution is to add a [multiline indicator](https://gist.github.com/rjattrill/7523554).

JoshData:
I think our file conforms to the YAML specification.

joshbetz:
JSON serialization is also built into most modern programming languages, while YAML is not.

essandess:
+1 to keep YAML as a raw format for ease of editing and correcting.

Also +1 to mirror a json for application speed because the YAML load time is ridiculously slow.

I've seen this myself with this lookup script [congress_lookup.py](../../../essandess/congress-legislators/blob/master/scripts/congress_lookup.py) to access contact information in the database discussed in #343.

essandess:
YAML load speeds are improved by 10x if the fast C Loader is used:

```python
[...]

PR #451, opened 2017-03-07
json / csv formats

joelcollinsdc:
- Add call alternate_bulk_formats from save_data
- add a travis command to try to ensure consistency
- update gitattributes to try to hide diffs from alternate_formats

Refs #362

WIP because needs review of how much the impact is for regular maintainers

edit: FWIW, making changes is a bit harder this way.  Since some of the scripts still require manual review and tweaks, we'll probably have to still manually run a script after hand editing the yamls but before committing...


legislators-current.yaml
```diff
@@ -39594,7 +39594,7 @@
     class: 2
     state_rank: junior
     party: Republican
-    url: http://www.rounds.senate.gov
+    url: https://www.rounds.senate.gov
```
konklone: @joelcollinsdc Did the data itself get updated in this PR? Noting a couple of substantive changes.

scripts/utils.py
```diff
@@ -188,8 +189,12 @@ def load_data(path):
...
 
 def save_data(data, path):
-  return yaml_dump(data, os.path.join(data_dir(), path))
-
+  yaml_dump(data, os.path.join(data_dir(), path))
+  
+  if (path=="legislators-historical.yaml"):
+    generate_alternate_formats(historical=True)
+  else:
+    generate_alternate_formats(historical=False)
```
konklone: Will this cause *all* data files to be regenerated if even one data file changes?

What I'd originally suggested is that when a specific data file is saved, that we save the CSV/JSON version of that data file right after we save the YAML version. I think the code would be simpler, and we could avoid a `historical` boolean flag, by taking that approach.

test/formats.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cd scripts
+python alternate_bulk_formats.py
+cd -
+if [ $(git diff alternate_formats/* | wc -l) -eq 0 ]; then
+  echo "No alternate format changes needed..."
+else
+  echo "Alternate formats need regenerated.  Run alternate_bulk_formats.py"
```
konklone: I do like having a sanity check script to make sure the files haven't gotten out of sync, but this approach won't actually catch bugs in `alternate_bulk_formats.py` itself, since it's using it to do the verification.

I think the ideal would be to have a script which, for each data file, opens up the YAML/CSV/JSON versions and does a "deep equal" approach to them to test for differences.

.gitattributes
```diff
@@ -0,0 +1 @@
+alternate_formats/* -diff
```
konklone: Could you say what this `.gitattributes` file is doing? I haven't used it before, and I did just go and try to learn more about it, but I'm having a hard time understanding what this line does.

joelcollinsdc: This was attempting to make the diffs of these files not show up, or not be as prominent at least, during local git commands.  

So you see this when doing a git diff, instead of the whole diff:
```
diff --git a/alternate_formats/legislators-current.json b/alternate_formats/legislators-current.json
index 9faeb22..956ccae 100644
Binary files a/alternate_formats/legislators-current.json and b/alternate_formats/legislators-current.json differ
```

legislators-current.yaml
joelcollinsdc: The last commit was an attempt to show what a sample change would look like with this approach.  I edited the top PR comment to indicate that this branch shouldn't be merged as is too, i can make a new branch / pr when we agree on an approach.

test/formats.sh
joelcollinsdc: The yaml/json deep comparison seems easy, thats a good idea.  The CSV part... since its not 1:1, should I just leave what is in alternate_bulk_formats alone, and separately create a json file for each yaml?  Or remove CSV from alternate_bulk_formats?

konklone: My vote would be to remove CSV from the set of supported bulk formats. We've seen strong community interest in JSON, but less so from CSV.

.gitattributes
joelcollinsdc: also @konklone, what are your thoughts on the json formatting being all one line.  Saves space (a few MB on the historical file...), and similar to the .gitattributes thing i was trying it makes looking at diffs on github a little easier (IMO, but obviously not looking at diffs within the json itself.  But in theory we can trust the diff matches if the build passes anyways)

konklone: If we're checking that the contents match the human-readable-and-diffable YAML file, that seems fine. We can optimize for performance on the JSON output.

joelcollinsdc:
Made the discussed changes

joelcollinsdc:
For some reason, for the committees-historical file, loading the yaml and json and going == doesn't seem to work.  (Works for the rest though...)  Not sure why.  Comparing keys and values showed no differences...

joelcollinsdc:
I force pushed a new commit with the test data removed.  The issue with comparing the historical committees file was due to json keys being converted to strings.  I think I added a workaround there.

test/test_json_matches.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
+
+import json
+import rtyaml
+import glob
+
+def jsonKeys2str(x):
+    """Some of the yamls have integer keys, which json converts to string.
+    in the future if there are keys that are strings that are intended to be left
+    as strings this may break"""
```
konklone: Note for later: I would not mind banning integer keys altogether and enforcing that every key is a string.

konklone:
@joelcollinsdc Sorry for letting this sit. I'm basically :+1: on it, but it's now not clear to me what role `alternate_bulk_formats` plays, and why it's still outputting CSV files. Can that script, and/or its CSV output lines, be removed?

JoshData:
I'm using `alternate_bulk_formats.py` and am committed to the output until I discontinue GovTrack's bulk data during Congress's summer recess.

konklone:
The tests work when I check out the branch, but they fail when the branch is merged into master (which is what he Travis build above shows). @joelcollinsdc I try to fix this by running `generate_json.py`, but this generates no changes (at least none that appear through `git diff`). Any ideas?

joelcollinsdc:
Are you saying that because the travis PR build is failing even though the push build is passing?  Yea I'm not sure what thats about.  Perhaps because the merge is taking legislators-current and social media from master and testing against the old json?  I think when we merge this we'd have to re-run generate_json...

konklone:
@joelcollinsdc That's what caused me to check it out, but I verified that when I actually checked out the branch, merged it against a copy of `master`, that the test script no longer worked. I then ran `generate_json.py` and that did not fix it or appear to change anything.

joelcollinsdc:
@konklone seems to have worked for me?  maybe i'm doing something different, or maybe its specific to my version of python...

konklone:
@joelcollinsdc When you re-ran the JSON generator, you fixed the build. :+1: from me.

joelcollinsdc:
Thanks @konklone .  Just wanted to point out that because #454 was merged in before this one, the json is not 100% up to date as a result...  I still think there would be less complication if we setup the mirrored repo that hosts updated json automatically...

PR #456, opened 2017-03-22
fix gillibrand and faso fax numbers

msimonborg:
Gillibrand's fax number was actually Faso's, fixed that and added Gillibrand's correct fax


konklone:
The failing build is from the JSON/YAML syncing we did in a prior PR. I tried running `generate_json.py` locally but it didn't generate any local changes. cc @joelcollinsdc

joelcollinsdc:
@konklone can you try removing the line in the .gitattributes that suppresses diffing the json files to see if that clears up confusion?  I tried removing it on my end, and now the diffs are massive... so I feel its actually serving a purpose, but...

PR #459, opened 2017-04-03
Twitter, youtube updates

pullyl:
Adding twitter handles for 3 reps
Adding many youtube ids


legislators-social-media.yaml
```diff
@@ -4166,6 +4175,7 @@
     twitter: RepSarbanes
     facebook: RepSarbanes
     twitter_id: 364415553
+    youtube_id: list=UUCypi0IDTr-XxlAXsJCdBZQ
```
konklone: This looks incorrectly formatted, as it begins with `list=`.

konklone:
These additionals all look great, except for the one `list=` mistake I noted. If you can fix that (or just delete the entry), this is :+1: from me. Did this come from running the `social_media.py` script in the repo?

pullyl:
thanks @konklone.  Not sure where that mistaken bit came from.

legislators-social-media.yaml
```diff
@@ -4175,7 +4175,7 @@
     twitter: RepSarbanes
     facebook: RepSarbanes
     twitter_id: 364415553
-    youtube_id: list=UUCypi0IDTr-XxlAXsJCdBZQ
+    youtube_id: UUCypi0IDTr-XxlAXsJCdBZQ
```
konklone: Hmm, actually this doesn't seem to resolve to a valid channel:

https://www.youtube.com/channel/UUCypi0IDTr-XxlAXsJCdBZQ

So perhaps it would be better to remove this one entirely, @pullyl.

konklone:
Thank you for making the changes and for your contributions here, @pullyl!

PR #464, opened 2017-04-26
Add Ron Estes

bycoffe:
Source: http://clerk.house.gov/floorsummary/floor.aspx?day=20170425

@JoshData if you know what his GovTrack ID will be, I can add it to this.


JoshData:
Was just working on this - will add govtrack momentarily. In the future you're welcome to assign the next unused integer to new people.

PR #468, opened 2017-05-03
Remove extra term for James Scheuer.

bycoffe:
He was not in Congress from 1973-1975 (he lost in the 1972 primary).
Source: http://bioguide.congress.gov/scripts/biodisplay.pl?index=S000124


dwillis:
Thanks!

PR #469, opened 2017-05-04
Remove duplicate Elmer Thomas entry.

bycoffe:
The entry with bioguide T000165 appears to be a duplicate of
the entry with bioguide T000177. There is in fact no bioguide
page for T000165. (There is a separate House history page --
http://history.house.gov/People/Detail/22830 -- that links to
the T000165 bioguide page, but the page doesn't exist.)


JoshData:
I can't believe we're still finding duplicates.... Thanks for spotting this. :+1:

dwillis:
Amazing.

PR #470, opened 2017-05-12
ran wikidata_bioguide.py

thinkcontext:
It added several Google, Vote Smart, and Ballotpedia ids.  Additionally it picked up a Wikipedia article name change for John Lewis.


dwillis:
Thanks @thinkcontext!

PR #474, opened 2017-05-25
add curly apostrophes to O', D', L' names

JoshData:
O'Rourke => O’Rourke (curly single quote), and so on for O’Halleran and a bunch of other historical people


bycoffe:
A few more names with apostrophes we might want to also change: `D'Alesandro`, `D'Ewart`, `D'Amato`, `L'Episcopo`, `D'Emory`, `D'Amours`, `L'Engle`.

JoshData:
Got it. Updated the PR in-place. I used a script to update apostrophes in all `name:` fields.

konklone:
Is this more grammatically correct in some way?

JoshData:
It's the more correct glyph. Straight apostrophes aren't really a thing outside of coding.

PR #482, opened 2017-06-26
use CircleCI to update gh-pages with latest downloadable files in YAML, JSON, CSV

JoshData:
For #480: Created a circle.yml file to use CircleCI to automatically push updated YAML, JSON, and CSV files to gh-pages. And added download links to README.

PR #484, opened 2017-06-27
run linter to fix build

joelcollinsdc:
Ran python lint.py


konklone:
Candidate for the CI build script?

JoshData:
Probably better to keep master as linted as possible? The CI script in #482 will stop if the files aren't linted with a failed test, as travis does.

Issue #372, opened 2016-10-22
FollowTheMoney Entity IDs?


bshor:
Hello,

[FollowTheMoney.org](http://FollowTheMoney.org) is a wonderful resource on campaign finance activity in all federal and state elections. It would be awesome if this project could incorporate its entity ids, which track a candidate's activities as recipient and donor across their political careers.

As an example, here's [Sherrod Brown's page](http://www.followthemoney.org/entity-details?eid=3439679), referenced by his entity ID.

JoshData:
I'm sure if someone did the work to match up the IDs, we'd gladly accept a pull request adding them. :)

showerst:
I found a list at https://github.com/followthemoneyorg/external-ids that has been recently updated, and has the openstates IDs. I'll try to write a script later that matches up as many as possible.

PR #487, opened 2017-07-01
Add NIMSP Entity IDs

showerst:
Issue #372 

Someone at votesmart [published a CSV](https://github.com/votesmart/political-id-match/) with FollowTheMoney (NIMSP) Entity IDs for many legislators, that also has their votesmart IDs. Since congress-legislators has votesmart IDs, I wrote a script to cross-reference and add them.

Note that this isn't my data, but I spot checked a number of them and they look good.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -643,6 +643,7 @@
     icpsr: 40700
     wikidata: Q22237
     google_entity_id: kg:/m/05fbpt
+    nimsp: '17659011'
   name:
     first: Amy
     middle: Jean
@@ -695,6 +696,7 @@
     icpsr: 40701
     wikidata: Q22260
     google_entity_id: kg:/m/040_w4
+    nimsp: '4742989'
   name:
     first: Claire
     last: McCaskill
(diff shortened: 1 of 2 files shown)


showerst:
Looks like this failed the test -- I can fix but just before I do:

1. Is this still a desired feature?
2. for the name, 'nimsp' or 'followthemoney' ? There's an old nimsp id too, so nimsp_entity might make sense too
3. Do I just add it to validate.py with type int, or is there another model/definition file somewhere that needs to be edited?

JoshData:
* Is this still a desired feature?

I don't know if anyone is actively looking to use this, but we might as well add it.

> for the name, 'nimsp' or 'followthemoney' ?

`followthemoney` would be more consistent with our `opensecrets` id (i.e. we don't call it the `crp` ID).

> Do I just add it to validate.py with type int, or is there another model/definition file somewhere that needs to be edited?

Right, just add it to validate.py. The values should be ints (unquoted) - they're currently strings.

You should also add it to the README. And since it looks like we don't have IDs for everyone, you should explain why in the README.

illuminates3:
Add to validate////××! M

PR #497, opened 2017-07-27
Added missing information

GoldenJoe:
Birthdates: Hoover, Eisenhower
Gender: Ford, Rockefeller

### DIFF
--- a/executive.yaml
+++ b/executive.yaml
@@ -987,6 +987,7 @@
     middle: Clark
     last: Hoover
   bio:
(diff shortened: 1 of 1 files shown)


executive.yaml
```diff
@@ -987,6 +987,7 @@
     middle: Clark
     last: Hoover
   bio:
+    birthday: '1874-8-10'
```
JoshData: There should be a leading zero in the month.

PR #504, opened 2017-08-24
Add social data

brennv:
Added twitter, facebook or instagram IDs to the social data associated with these bioguide IDs:
`A000367, C001049, C001075, C001108, P000603, K000367`

Added legislators, and twitter or facebook IDs for:
`G000584, H001078, N000190, P000258`

### DIFF
--- a/legislators-social-media.yaml
+++ b/legislators-social-media.yaml
@@ -5408,3 +5408,10 @@
     twitter: RepJimmyGomez
     twitter_id: 2371339658
     facebook: RepJimmyGomez
+- id:
+    bioguide: N000190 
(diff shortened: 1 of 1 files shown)


JoshData:
Hi. Per the README, we "include accounts that are paid for with public funds and which represent official communications of their office." We don't include accounts managed by their campaigns. Could you review your additions and check whether the accounts are associated with their official office or their campaign? Official accounts should be linked from their official website and when they're not we call up their congressional office and ask.

brennv:
Reverted, for now it looks like only updates for `N000190` are backed by an official site.

PR #507, opened 2017-08-24
Fix Twitter IDs in YAML-JSON conversion

alexlitel:


JoshData:
Looks ok. I'd prefer a little more code comments so we remember why this code block is there. Mind describing the issue in there?

alexlitel:
Added some comments that should hopefully resolve any potential future confusion.

JoshData:
Thanks!!

PR #509, opened 2017-08-28
Update social_media_blacklist.csv

meghjoshi123:


konklone:
This looks not intended. Are you trying to change the blacklist in some way?

PR #510, opened 2017-09-01
ran wikidata_bioguide.py

thinkcontext:
It picked up a bunch of Wikipedia renames as well as a smaller number of Vote Smart and Ballotpedia ids.  Made a minor error checking adjustment to the script.


JoshData:
Hey. Could you just remove the `alternate_formats` file from the PR? (These are now gone from master and are added automatically in the gh-pages branch.)

JoshData:
Thanks!

Issue #511, opened 2017-09-03
Data lists Paul Ryan as chair of the Joint Committee on Taxation


konklone:
But Kevin Brady is chair:

https://www.jct.gov/

I'm a little out of touch with how we get House memberships of joint committees, but could refresh myself unless someone's already pretty familiar with how that might have happened.

PR #513, opened 2017-09-20
Joint Committee on Taxation Membership

Panpog1:
Fixes issue #511

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -12446,12 +12446,6 @@ JSTX:
   thomas: '01531'
   bioguide: S000770
   chamber: senate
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks. This only gets us part way there though. According to https://www.jct.gov/, Paul Ryan isn't on the committee, so we'd want to remove him, and if we can vet the rest of the membership while we're here that'd be great.

Panpog1:
Members still need to be added. I'll look into that.

Panpog1:
I removed nonmembers.

JoshData:
HI there. So.... jct.gov now says Brady is the Vice Chair....

PR #515, opened 2017-10-05
Tim Murphy to resign on Oct 21

JoshData:
Just queuing this up to merge on Oct 22.


konklone:
:+1: from me. For reference, the build error is that the retirement entry `has an end date in the future but is in the historical file.`, which is correct. It'd be nice to kick the build to run again before merging so that it's clean.

JoshData:
Great.

Just want to note let's NOT merge until after he actually resigns, so on the 22nd (i.e. assuming the resignation is COB of the 21st). :)

konklone:
Yep, that was my assumption, and was just noting that for when it is finally merge-time.

bycoffe:
Here's his resignation letter in the Congressional Record; it was submitted yesterday. Just says "effective October 21": https://www.congress.gov/congressional-record/2017/10/10/house-section/article/H7895-7

PR #516, opened 2017-10-10
Combine the first and middle name fields when the first name is just an initial

JoshData:
On GovTrack, the logic for constructing legislator display names for FIRST LAST or LAST, FIRST format has always been:

1. take the first name

2. if the first name ends with a period (i.e. is an initial or in Wm. Lacy Clay's case, an abbreviation), forget the first name and use the middle name instead, and in other cases ignore the middle name

This logic has worked well and generally matches the display form that the legislator uses themselves. For instance, we have `first: F.` / ` middle: James` for Rep. Sensenbrenner, who generally goes not by `F. ` or `F. James` but by... `Jim`:

image

 except in some cases where the first initial is sometimes also used, like "F. James Sensenbrenner":

image

When the first name ends in a dot, the middle name seems to always be obligatorily a part of the proper way of displaying the name when showing a first name, either alone or with the first initial. In other cases, when the first name does not end in a dot, the middle name is generally _not_ used for display purposes at all.

This commit moves the middle name value to inside the first name field (with a preceding space) just when the first name ends in a dot, the purpose of which is to make it easier for users of this dataset to correctly construct display names for legislators. The new logic would be:

1. use the first name

2. WIN (i.e., don't ever use the middle name field)

This would differ from GovTrack's current logic by always including first initials instead of dropping them, which may not be better but is probably equally ok.

What do folks think?

Other alternatives to consider would be

* completely drop middle name values that are generally not included when displaying a legislator's name so that always including the middle name value would yield a more correct display string (than displaying all of the middle names we have now)

* move these middle names to the first name (i.e. `middle: Clay` => `first: Clay`) and move first initials to a new `first_initial` field (`first: Wm.` => `first_initial: Wm.`) so that data users would be more likely to correctly use the middle names in these cases and could choose whether or not to display the first initial

* add flags that indicate how to use the first/middle name fields in good display strings, i.e. `use_first: False` `use_middle: True` to suggest friendly name strings like "Lacy Clay" which is how the congressman is generally known

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,8 @@ The following fields are available in `legislators-current.yaml` and `legislator
 	* bioguide_previous: When bioguide.congress.gov mistakenly listed a legislator under multiple IDs, this field is a *list* of alternative IDs. (This often ocurred for women who changed their name.) The IDs in this list probably were removed from bioguide.congress.gov but might still be in use in the wild.
 
 * name
-	* first: The legislator's first name. Sometimes a first initial and period (e.g. in W. Todd Akin), in which case it is suggested to not use the first name for display purposes.
-	* middle: The legislator's middle name or middle initial (with period).
+	* first: The legislator's given name. When a legislator uses a first initial and their middle name, this field combines both as a single string value with a space.
+	* middle: The legislator's middle name or middle initial (with period). The middle name is generally not used when displaying a legislators name. When it is, it is included in the `first` name field.
 	* last: The legislator's last name. Many last names include non-ASCII characters. When building search systems, it is advised to index both the raw value as well as a value with extended characters replaced with their ASCII equivalents (in Python that's: u"".join(c for c in unicodedata.normalize('NFKD', lastname) if not unicodedata.combining(c))).
 	* suffix: A suffix on the legislator's name, such as "Jr." or "III".
 	* nickname: The legislator's nick name when used as a common alternative to his first name.
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -6384,8 +6384,7 @@
     wikidata: Q959455
     google_entity_id: kg:/m/024vjr
   name:
-    first: Wm.
-    middle: Lacy
+    first: Wm. Lacy
     last: Clay
     suffix: Jr.
     official_full: Wm. Lacy Clay
@@ -7000,8 +6999,7 @@
     wikidata: Q538944
     google_entity_id: kg:/m/04cmt7
   name:
-    first: K.
-    middle: Michael
+    first: K. Michael
     last: Conaway
     official_full: K. Michael Conaway
   bio:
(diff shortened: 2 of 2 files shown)


bycoffe:
I think each user is going to have a different idea of what form of a name they want to display, and keeping first/middle/last separate makes it more flexible.

I don't think having a hard and fast rule about combining first initial and middle name is going to work for all cases; "F. James Sensenbrenner" seems better than "James Sensenbrenner", but "J. Robert Kerrey" seems worse than "Robert Kerrey", and "T. Slade Gorton" worse than "Slade Gorton".

Of the alternatives you mentioned, the third seems best to me, though maybe there are others that would work well.

konklone:
The way we handled this at Sunlight was to have a separate field: `nickname`, that could be used to reliably display the person's name.

So you might have someone with the fields:

```yaml
first_name: F.
middle_name: James
last_name: Sensenbrenner
nickname: Jim
```

And for someone with no special nickname, the field is just blank or missing:

```yaml
first_name: John
middle_name: Sidney
last_name: McCain
nickname: 
```

This made display logic very easy, at the cost of manually maintaining nickname fields. But the cost of maintaining nickname fields is actually super low -- they're not used by everyone, they basically never change over the course of someone's life (and they're not traditionally affected by marriage like last names can be). So you do them once at election-time, or whenever you eventually notice it, and it flows naturally downstream.

So display logic can be very consistent and simple for those who need it:

* If there's a nickname field, use it
* If not, use the first name field

And I think it gets rid of all the tradeoffs in the approach and alternatives given. Thoughts?

JoshData:
We already have a nickname field, but it often holds an informal name (e.g. Bernie) that wouldn't be appropriate for my use (Sanders goes by Bernie on his website, but the Senate refers to him as Bernard). There are other clearer cases where the nickname wouldn't be appropriate to replace the first name (Pete King, Chuck Schumer, Bob Casey, Jimmy Duncan -- these are not appropriate in a formal setting). On GovTrack I display nicknames in quotes (Sen. Bernard “Bernie” Sanders). So we'd need a new field or we'd need to purge the informal nicknames from our data.

konklone:
> (Pete King, Chuck Schumer, Bob Casey, Jimmy Duncan -- these are not appropriate in a formal setting)

Are you sure? I've always been comfortable displaying those names to users, no matter the setting, as it's how users will recognize those names. Bernie Sanders doesn't go by "Bernard Sanders" in either personal or professional settings, even if Senate processes and information systems choose to use his legal name. At formal events in the Senate, I'm confident he introduces himself, and is generally introduced as, Bernie -- even if the Senate presiding officer might say "Sanders, Bernard" during roll call.

I think of the Senate's use of legal names as similar to an HR or IT department requiring that Bernie's email be `bernard.sanders@senate.gov` -- a functional thing, but not about formality in a personally meaningful way.

JoshData:
> Are you sure?

On the whole, yes.

JoshData:
Closing because there was no resolution here. Thanks all.

PR #518, opened 2017-10-19
Add district offices

tcarobruce:
This addresses #343 by making the following changes:

- adds the `legislators-district-offices.yaml` file
- documents the file and its fields in `README.md`
- extends `scripts/alternate_bulk_formats.py` to produce corresponding JSON and CSV files
- does validation of the district offices file using `scripts/office_validator.py`
- modifies `test/validate.py` to do office validation (for TravisCI/CircleCI)

The helper script used to do geocoding is also included. The push to gh-pages should work as well, though I'm unable to test that.


JoshData:
Looks good to me. Thanks! Any other maintainers have thoughts?

I won't block a merge with these comments but just some data thoughts:

* It's normally a little cleaner to omit fields that have empty values, i.e. drop lines entirely whose values are the empty string.
* A more logical field order would make the data easier to read, e.g. id first, then all the address fields in the order they would appear on an envelope, then phone and fax, then longitude/latitude.

tcarobruce:
I like both of these ideas, made the changes and added (non-error) checks to the validation script. Lat/lon are included after address since they're geographical.

konklone:
@tcarobruce Thanks for the change! That looks good.

It looks like the district office validation script is in fact creating some build errors:
https://travis-ci.org/unitedstates/congress-legislators/builds/292335968

Any thoughts on those errors?

tcarobruce:
Tim Murphy (M001151) resigned in the interim. I merged in the upstream change and removed his offices. Looks like it's working now!

Perhaps the office validator should suppress warnings when run under CI to make errors easier to find.

konklone:
> Perhaps the office validator should suppress warnings when run under CI to make errors easier to find.

Yeah, that does sound good -- it appeared to be a deluge of errors at a quick glance.

But not a blocker for this PR, IMO. @JoshData, this is good with me if it's good with you.

JoshData:
Looks good to me!

konklone:
:tada:

PR #519, opened 2017-10-27
District office fixes

tcarobruce:
Small fixes to district office scripts:

- validator option to suppress warnings (used by CI validate script to reduce noise)
- add header row to district office CSV
- minor fix for validator script to prevent erroring on unknown fields


konklone:
Thank you!

PR #521, opened 2017-11-08
Update offices for Rep. Gianforte

tcarobruce:
Verified changes from house.gov site.


konklone:
Thanks, @tcarobruce. If you can drop the URL of the page in future PRs from where you got the info, that'd be helpful, but much appreciated either way.

Issue #523, opened 2017-11-13
Adding new members without district offices


bycoffe:
I'm trying to add John Curtis, R-UT, to the `legislators-current.yaml` file, since he's being sworn in today. He doesn't appear to have any district offices yet. Without an entry (including an office address) in `legislator-district-offices.yaml`, the `validate.py` test fails.

I'm not sure what the best way to handle this is. It seems like we should be able to list new legislators as soon as they're sworn in, even if we don't know the address of a district office. One option might be to include them in the `legislator-district-offices.yaml` file, but without any offices (right now doing that also results in an error). Another might be to make missing offices a warning rather than an error.

Does anyone have thoughts about this? Or is there a way to handle it already and I'm just missing it?

JoshData:
Thanks for getting on this. What about looking at the swearing-in date. We can make it a warning if it's two months (?) after swearing-in without district offices?

bycoffe:
Sounds good to me -- not ideal that whoever makes the first change after two months will see an error even if it's unrelated to their change. But we can encourage whoever adds the member in the first place to keep an eye out for district offices and add when available.

Here's a pull request with that change: https://github.com/unitedstates/congress-legislators/pull/524

PR #524, opened 2017-11-13
Allow new members without district offices

bycoffe:
This only reports an error if the member has been in office for at least 60 days. See https://github.com/unitedstates/congress-legislators/issues/523


JoshData:
Can we make it a warning otherwise? (Does a warning make the tests pass?)

bycoffe:
I changed it to a warning .. it'll now print the warning but exit with code 0 (rather than 1 for an error)

JoshData:
Thanks!

PR #533, opened 2017-12-09
correct party of Ander Crenshaw in 2001-2002

JoshData:
Not sure where this error came from - it has been in the data since the beginning.


konklone:
Noting for reference here:
http://bioguide.congress.gov/scripts/biodisplay.pl?index=C001045

JoshData:
Thanks!

PR #537, opened 2018-01-02
Al Franken resigns effective 2018-01-02

bycoffe:
https://www.mprnews.org/story/2018/01/02/al-franken-resigns
http://www.startribune.com/transition-time-franken-moves-out-smith-moves-in/467678623/

PR #539, opened 2018-01-03
Add Doug Jones, Tina Smith; retire Luther Strange

JoshData:
This PR is based on @bycoffe's commit bycoffe/congress-legislators@56b2c10 per #537.

I zapped Jones's and Smith's office/phone/contact_form fields since I'm not sure if they'll be the same as their predecessor? (Smith's contact_form had franken.senate.gov so I had to zap at least that.)

As @bycoffe mentioned it won't validate until tomorrow. Jones and Smith will be sworn in in the afternoon after the 2nd session starts, so I guess we could wait until the afternoon before merging. Or we could merge tonight ahead of time.


JoshData:
Doug Jones has the wrong party. Will fix...

bycoffe:
Thanks, sorry about that

JoshData:
PR updated. Could someone review and merge?

JoshData:
Oh, also- I changed Jones's name to Doug (vs Gordon Doug) because his FEC registration only used Doug and everyone knows him as Doug, so I think Doug fits better. That ok?

bycoffe:
Fine with me.

PR #540, opened 2018-01-03
SenKaineOffice (twitter) seems to have been deleted

JoshData:
The account isn't linked from his homepage anymore and doesn't exist on twitter. I think the numeric ID is invalid now so that means the account was deleted/inactivated and not renamed.


JoshData:
Without objection it is so ordered.

PR #541, opened 2018-01-12
add MapLight and OpenSecrets IDs for new legislators

mrumsky:
add MapLight IDs for John Curtis, Doug Jones, Tina Smith
add OpenSecrets ID for Doug Jones


JoshData:
Thanks!

PR #546, opened 2018-02-11
Update committee membership

justinvasel:
- Some chair/ranking member details have changed since the last time it was run. One example was the HSBU, whose ranking member before this update was listed as someone who is now a Senator.
- For some reason the Thomas codes appeared in a different position than before, so this commit looks way more substantial than it really is.
- There were several sanity check warnings about name mismatches, which were in regard to punctuation and a middle initial. Full output:
    ```
    Name mismatch: John Curtis (in our file) vs John R. Curtis (on the Clerk page)
    Name mismatch: A. Drew Ferguson IV (in our file) vs A. Drew Ferguson, IV (on the Clerk page)
    Name mismatch: A. Drew Ferguson IV (in our file) vs A. Drew Ferguson, IV (on the Clerk page)
    Name mismatch: A. Drew Ferguson IV (in our file) vs A. Drew Ferguson, IV (on the Clerk page)
    Name mismatch: A. Drew Ferguson IV (in our file) vs A. Drew Ferguson, IV (on the Clerk page)
    ```
    Wasn't sure if I should address these by manually updating the yaml, but if so let me know and I'll add that commit to this PR.

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -3,428 +3,428 @@ HLIG:
   party: majority
   rank: 1
   title: Chair
-  thomas: '01710'
   bioguide: N000181
+  thomas: '01710'
 - name: K. Michael Conaway
   party: majority
   rank: 2
-  thomas: '01805'
   bioguide: C001062
+  thomas: '01805'
 - name: Peter T. King
   party: majority
   rank: 3
-  thomas: '00635'
   bioguide: K000210
+  thomas: '00635'
 - name: Frank A. LoBiondo
   party: majority
   rank: 4
-  thomas: '00699'
   bioguide: L000554
+  thomas: '00699'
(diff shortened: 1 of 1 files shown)


JoshData:
Hey, thanks for this!

I'm not sure why the thomas/bioguide ID fields switched order. It happens when I run committee_membership.py too --- that's new. I'm opening a new PR #547 that makes a fix to the script to ensure the order comes out right, and then I also included in my PR the same update to the YAML file that you did but with the corrected script.

So I'm going to close this PR and we'll merge the other one. You did the right thing - our script had a bug - and we'd definitely be glad to have you do exactly this in the future if you notice committee membership has gone out of date. :)

justinvasel:
Great! Thanks for the quick response.

PR #547, opened 2018-02-11
committee membership ID order

JoshData:
For some reason, running committee_membership.py now suddenly reverses the thomas and bioguide ID fields compared to the order that we have in the YAML file, creating spurious diffs.

We were using a dict to pass this information, and dicts have been unordered (though that changes in CPython 3.6, though I'm running 3.5 so that wasn't the problem), so the order was only lucky to be stable until now. It's surprising we got this far without seeing them flip.

Changing it to an OrderedDict should guarantee the order that we want.

This PR also updates the committee-membership-current.yaml data file after running the script.
  
See #546.

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -2060,129 +2060,134 @@ HSAS:
   bioguide: R000575
 - name: Bill Shuster
   party: majority
-  rank: 9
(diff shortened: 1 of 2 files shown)


konklone:
Good catch!

PR #549, opened 2018-02-14
Doug Jones DC and District offices

jlev:
Offices sourced from https://www.jones.senate.gov/content/contact-senator
Mobile and Montgomery offices do not list phone contacts
Geocoded with `python geocode_offices.py`


konklone:
:+1:, looks great. Thanks for catching this!

PR #551, opened 2018-02-22
Added senate_class field to be in CSV files

dannguyen:
This is derived from the YAML


JoshData:
Thanks! We'll see if it works momentarily when CircleCI generates the files....

PR #552, opened 2018-02-25
mark Sen. Sanders as having always caucused with the Democrats

JoshData:
Recently-ish we added the `caucus` field to indicate the party Independents caucus with. We currently have it set for Sen. Sanders's current term, Sen. King (who is in his first term), and MP's delegate Rep. Sablan's previous term (he's currently a Democrat but was an Independent).

This PR sets it for all of Sanders's previous terms so that his party+caucus affiliation is consistent for all of his terms, and it makes it so all terms in legislators-current with `party: Independent` have `caucus:` set.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -964,42 +964,49 @@
     state: VT
     district: 0
     party: Independent
+    caucus: Democrat
   - type: rep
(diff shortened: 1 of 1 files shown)


JoshData:
There being no objection, merging...

PR #555, opened 2018-03-16
Add district offices for Sen. Tina Smith (MN)

justinvasel:
Add all four district offices for Tina Smith (Moorhead, Duluth, Rochester, St. Paul).

The lack of these entries was causing `test/validate.py` to fail.


alexlitel:
Can someone verify and resolve conflicts with this update?

konklone:
Confirmed: https://www.smith.senate.gov/

screenshot from 2018-04-22 12-31-00

PR #556, opened 2018-03-16
Rep. Slaughter passed away some time between last night and this morning

JoshData:
I marked the term end date as yesterday. [Bioguide](http://bioguide.congress.gov/scripts/biodisplay.pl?index=S000480) (already) has the death date as today, but since it was at least early A.M. I thought it made more sense to indicate yesterday as the last date.


bycoffe:
For at least the past several members who died in office, we've used the death date as the end of their last term. Shouldn't we do the same here?

dwillis:
I'm with @bycoffe on this.

JoshData:
Fine by me, updated.

AlJohri:
hi, are the death dates parsed out of the bioguide anywhere?

bycoffe:
@AlJohri this script will get death dates from wikidata: https://gist.github.com/bycoffe/814fe1a277327080101ffc5451c27232

AlJohri:
@bycoffe wow, that was fast! thanks!

PR #567, opened 2018-05-08
added Rep. Lesko

JoshData:
A bioguide ID isn't yet posted.


dwillis:
House clerk has it as L000589

JoshData:
Updated with bioguide ID.

Issue #576, opened 2018-06-21
How should P.O. box only offices be represented?


csnardi:
In #574, I updated `G000582-san_juan` from a physical location to one that only represents a P.O box per https://gonzalez-colon.house.gov/contact/offices/san-juan-office. As such, I removed the geocoded location, since the lat/long would be inaccurate. However, this no longer passes the validator, as there is an address but no lat/long. How should we represent this? I note that 3 offices that are P.O. box only are indicated with a `suite`, thus not making that rule of the validator be triggered. All others have lat/longs, although it's unclear as to what those point to.

tcarobruce:
I propose removing (or leaving null) latitude and longitude for offices that list P.O. Boxes or are telephone-only, and removing lat/long from office validation.

PR #582, opened 2018-07-06
Don't error on missing lat-lon for district office

tcarobruce:
This solves #576, while still warning if latitude and longitude are missing from offices with addresses.


tcarobruce:
Sorry, picked up some unexpected merge data, will close and re-request.

PR #583, opened 2018-07-06
Don't error on missing latitude/longitude

### LINKED ISSUE #576: How should P.O. box only offices be represented?

tcarobruce:
Fixes #576, but still warns for offices with address but missing latitude and longitude.

### DIFF
--- a/scripts/office_validator.py
+++ b/scripts/office_validator.py
@@ -132,7 +132,7 @@ def check_legislator_offices(legislator_offices, legislator):
 
         if (office.get('address') and
                 not (office.get('latitude') and office.get('longitude'))):
-            errors.append("Office %s needs geocoding" % office_id)
+            warnings.append("Office %s missing geocode" % office_id)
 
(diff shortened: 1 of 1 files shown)


jsfenfen:
@JoshData fwiw I'm really interested in knowing all the mailing addresses, even PO boxes, because they sometimes show up in other records--and PO boxes seem vaguely sketchier for no good reason--but am not doing anything in particular with it right now.

JoshData:
Ah ok!

PR #597, opened 2018-09-05
Jon Kyl might be sworn in today

JoshData:
If Kyl is sworn in today, here is that. I assumed he'd be using McCain's office so I copied forward that address. I also assumed Flake is now and would still be the senior senator.


JoshData:
He was just sworn in:

https://twitter.com/greggiroux/status/1037384598051799040

Issue #41, opened 2013-03-05
Indicate whether senate (originally: senate & executive) term is elected, appointed, or something else


dwillis:
Would cover cases like LBJ's first "term" and senate appointments to fill vacancies.

[UPDATE: Executive is done. Just senate left.]

GPHemsley:
Was this work ever completed, or is this still pending?

PR #599, opened 2018-09-05
Mark Senate appointments and upcoming special elections

JoshData:
This PR makes two changes:

* Adds `how: appointment` to all senators appointed to fill a vacancy since 1913 (when that became possible) and updates start dates that appear to be erroneous, thanks to https://www.senate.gov/senators/AppointedSenators.htm. Where the appointment date was after our term start date, the term start date is updated. (In other cases I presume our term start date to correctly reflect the swearing-in date or the start of the session.) Where there was ambiguity on the Senate page about appointment date versus effective date, I used the effective date. See #41.
* For senators currently serving per an appointment to fill a vacancy, and when a special election will take place prior to the end of the term (which is the case for all current senators serving by appointment), I've changed the term end date to the future expected special election date. This makes the end date a better reflection of the actual expected end date of the term. I also set set `end-type: special-election` on these terms so that we can distinguish end dates in the future that represent the constitutional end of the term and end dates in the future that are expected special election dates. Note though that when a special election occurs, the actual end date of the term will typically be (and will later be revised to be) some time after the special election, because as far as I understand the term ends when the new senator is sworn in, which would be after the election date.


JoshData:
Last call before I self-merge....

Issue #605, opened 2018-09-24
Better name handling for CSV Export


joearasin:
For legislators who don't go by their first name (i.e. Jim Sensenbrenner, Lou Correa), the current name handling in the CSV is insufficient (as only first and last are included)

JoshData:
I think we'd all be fine with adding more columns if you want to submit a pull request. :)

PR #606, opened 2018-09-25
Add name fields to CSV export

joearasin:
Add additional name fields to CSV.

Addresses #605.


JoshData:
Thanks! The files linked from the README should update automatically momentarily....

PR #611, opened 2018-10-09
Charles E. "Chuck" Schumer

dphiffer:
Pretty small change to the Chuck Schumer `full_name`.

Before: `Charles E. Schumer`
After: `Charles E. "Chuck" Schumer`

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -24975,7 +24975,7 @@
     middle: E.
     nickname: Chuck
     last: Schumer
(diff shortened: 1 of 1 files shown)


JoshData:
This field is automatically maintained by the `senate_contacts.py` script, so we can't make manual modifications, though we can change the script if it's desired.

dphiffer:
Ah, fair enough. This is something I can take care of downstream. Thanks @JoshData!

PR #614, opened 2018-11-07
Nov 2018 Special Election Results - Smith, Morelle, Hern

JoshData:
I've started a branch (and this PR) for special election results.

According to https://ballotpedia.org/Special_elections_to_the_115th_United_States_Congress_(2017-2018) and other sources there were six special elections plus one election (OK-1) that included simultaneously filling a vacancy:

| Contest | Result |
| --------- | -------- |
| MN (senate) | Smith |
| MS (senate) | runoff |
| MI-13 | Jones |
| NY-25 | Morelle |
| OK-1 | Hern |
| PA-7 | no data on CNN |
| PA-15 | no data on CNN |

Sen. Smith (MN) won reelection and so the one commit in this PR adds a term for her. I set the end/start dates to Nov 6/7 for now. I imagine she will be sworn in again later this month, and in that case we should update the turnover dates to that date, I think, and then merge the change, maybe?

The MI-13 and NY-25 results seem to be in, but since they're new members I haven't done anything for them yet, and we'll wait until they're sworn in before we make the change anyway.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -41112,3 +41112,47 @@
     address: 1203 Longworth House Office Building; Washington DC 20515-3512
     office: 1203 Longworth House Office Building
     phone: 202-225-5355
+- id:
+    bioguide: H001082
+    fec:
+    - H8OK01157
+    govtrack: 412748
+    wikipedia: Kevin Hern
+    wikidata: Q58333614
+    ballotpedia: Kevin Hern
+  name:
+    first: Kevin
+    middle: R.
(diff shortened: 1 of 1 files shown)


JoshData:
Updated the PR:

* Revised Sen. Hude-Smith's end date to be the runoff date so the end date remains in the future.
* Added Reps. Hern and Morelle who are being sworn in today. https://twitter.com/CraigCaplan/status/1062320387277705217

I'd like to merge this today.

PR #623, opened 2018-12-10
Update historical committees and yaml file

### LINKED ISSUE #612: Path forward for historical committees
https://github.com/unitedstates/congress-legislators/blob/master/committees-historical.yaml hasn't been updated in two years, since thomas was turned off.

It's now missing some subcommittees like the [Subcommittee on Government Operations](https://oversight.house.gov/subcommittee/government-operations/) under the House Committee on Oversight and Government Reform 

Should I just manually add this subcommittee to historical committees file? Is there some more authoritative source of committees and subcommittees (there seemed to be some conversation of this in #128)

If I manually add subcommittees, what should I put for a thomas ID? Or is that an optional field?

fgregg:
This PR uses govinfo instead of the defunct thomas to get historical committee information.

Closes #612

### DIFF
--- a/committees-historical.yaml
+++ b/committees-historical.yaml
@@ -423,6 +423,7 @@
   - 112
   - 113
   - 114
+  - 115
   names:
     93: Standards of Official Conduct
     94: Standards of Official Conduct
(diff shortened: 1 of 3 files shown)


fgregg:
The broken CI is not related to this PR.

PR #624, opened 2018-12-15
Kyl, Lujan Grisham to resign at the end of the year

JoshData:
This PR should be merged once Kyl has actually left office. If Kyl's actual resignation date turns out to be different, this PR will need to be revised.


JoshData:
Updated with Rep. Lujan Grisham's resignation at the end of the year as well.

PR #625, opened 2018-12-15
The 116th Congress, fixes #615

### LINKED ISSUE #615: The 116th Congress
Ok. So. We just had an election. See #223 and #375 for how we've handled election results for the past two Congress. Yesterday's special elections are being done in #614.

Yesterday's election was for all representatives (incl. delegates) except Puerto Rico's delegate which is a four-year position and this was an off-year, for a total of 440 seats, and the 33 class 1 senators. The House Clerk promises to provide XML data mid November with bioguide IDs provided for all representatives, so we should use that wherever possible to automate this. I think we'll need to gather Senate results manually.

Cribbing from #375, the overall tasks that need to be done are:

As data comes in:
- [ ] Incumbents that won reelection: Clone their term but update the dates. Check if Senate state junior/senior rank needs to be updated.
- [ ] Legislators that won election to a new seat, e.g. representatives who ran for senate or who changed district number (possibly in PA where the districts were redrawn): Add a term with new information. Senate terms are likely junior rank.
- [ ] Incumbents and legislators who ran for a different seat but lost: Run the `retire` and `sweep` scripts to move them to the legislators-historical file and clear their committee and social media records.
- [ ] Winners who served in Congress previously: Move their record from the legislators-historical file to the legislators-current file and check if any missing metadata can be filled in. Senators are likely junior rank.
- [ ] Winners who are new: Add per the [new member template](https://github.com/unitedstates/congress-legislators/blob/master/misc/new-member-template.yaml). Senators are likely junior rank. We should have all representative bioguide IDs soon but senator bioguide IDs may not be available until they are sworn in.
- [ ] Add end dates to all current `leadership_roles` for Members who are no longer serving.
- [ ] Committee assignments must be cleared.
- [ ] [Photos](https://github.com/unitedstates/images) need to be found for new members.
- [ ] Social media accounts can be added for new members.

After January 3:
- [ ] Fix start dates to actual swearing in dates if any legislators didn't get sworn in on the first day.
- [ ] Bioguide IDs must be filled in when they become available. We have a script to guess IDs. We should know all representative IDs by now but not Senate IDs.
- [ ] New leadership roles
- [ ] Fill in the inevitable missing data, like office addresses (via the `house_contacts.py` and `senate_contacts.py` scripts) and committee membership.

JoshData:
This is a first pass at updating the legislator data for the 116th Congress following the 2018 regular general elections.

The changes are done entirely by script from the House Clerk 116th Congress preliminary data and a spreadsheet put together by GovTrack staff on senate election results.

There are two House seats whose elections aren't over, so this isn't final.

And of course we won't merge this until January 3 when the terms actually begin.


JoshData:
Fyi I just force-pushed the branch so that it will cleanly merge with master tomorrow. (I had messed up an earlier commit and it was totally unmergable.)

mikeshyavitzcbo:
Hey josh - Thanks for your work on this. Still not seeing any members in 116th... You still merging those changes?

PR #628, opened 2019-01-03
Fix indentation of FEC ID for lint

csnardi:
The lint didn't like the extra indentation for Hyde-Smith's FEC ID, so remove it.


csnardi:
Seems to be fixed in #625.

PR #629, opened 2019-01-03
Don't require build success for deploy

csnardi:
Oftentimes, the lint will fail for some reason or another, but we should still deploy the latest changes to GitHub pages. Remove the requirement that the build step must succeed for deploy to occur.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
       - image: circleci/python:3.5-browsers
     steps:
       # Set up.
-      - checkout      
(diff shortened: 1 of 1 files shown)


JoshData:
We might want to revert in the future to help keep the data clean, but for now this makes sense thanks!!

mikeshyavitzcbo:
Data! Thanks, @JoshData!

csnardi:
Agreed, I think the CircleCI check would need to be enforced on PRs first since this breakage has happened multiple times before.

PR #633, opened 2019-01-03
Official accounts for 14 Freshmen

frisch1:
These came from a crawl of *.house.gov pages of each, and were then hand verified for a "checkmark" on twitter, along with some representation of the official account. They are on the bottom...

### DIFF
--- a/legislators-social-media.yaml
+++ b/legislators-social-media.yaml
@@ -4482,3 +4482,69 @@
   social:
     facebook: RepGianforte
     youtube_id: UCj4ocM2PHNKBbmMQF8k-wsQ
+- id:
+    bioguide: B001306
+    govtrack: 412747
+  social:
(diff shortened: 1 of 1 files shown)


legislators-social-media.yaml
```diff
@@ -4482,3 +4482,69 @@
...
+    twitter: RepBalderson
+- id:
+    bioguide: A000376
+  social:
+    twitter: RepColinAllred
+- id:
+    bioguide: B001306
+    govtrack: 412747
+  social:
+    twitter: RepBalderson
```
csnardi: There's a duplicate entry here for Balderson, but this otherwise looks good.

JoshData: Good catch.

JoshData:
Thanks & merged! I removed the duplicate entry in the merge commit.

PR #636, opened 2019-01-04
Fix two Senate URLs

csnardi:
Two new Senators accidentally had their old House URLs put in for their Senate term. Fix these two URLs to point to their new Senate URLs.


JoshData:
Thanks!

PR #641, opened 2019-01-04
116th legislator wikipedia ids

dphiffer:
Pretty much like you'd expect: added `wikipedia` IDs for all the incoming congresspeople I could find on Wikipedia.

Example:

```yaml
- id:
    bioguide: D000630
    fec:
    - H8NY19181
    govtrack: 412805
    wikipedia: Antonio Delgado (politician)
  name:
    first: Antonio
    last: Delgado
```

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -38032,6 +38032,7 @@
     fec:
     - H8AZ09040
     govtrack: 412753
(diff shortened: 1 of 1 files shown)


JoshData:
Looks great thanks!

PR #642, opened 2019-01-05
rename the Oversight and Ed and Labor committees per title 1 of the 116th Congress House rules package H.Res. 6

JoshData:


JoshData:
Updated to change the names in the `jurisdiction` fields too.

JoshData:
Without objection it is so ordered.

PR #650, opened 2019-01-09
SenBennetCo twitter account was renamed, RepIlhanOmar disappeared

JoshData:
Not sure if RepIlhanOmar was never the right name or if it was made private or deleted.


dwillis:
@JoshData Interesting. `RepIlhanOmar` definitely existed yesterday, but had not tweeted.

PR #652, opened 2019-01-11
Update historical committees based on domain expert edits

fgregg:
@JoshData, a colleague at the [Lugar Center](http://www.thelugarcenter.org/) has suggested the following edits to the historical committees based upon her expertise (she works on committee oversight).


fgregg:
Sorry not ready yet.

PR #655, opened 2019-01-23
Rep. Tom Marino resigned, effective 2019-01-23

bycoffe:
http://clerk.house.gov/floorsummary/floor.aspx?day=20190123&today=20190123


JoshData:
Thanks!

Issue #492, opened 2017-07-12
Robert Menendez - Last name inconsistency


ezoehunt:
Robert Menendez (bioguide M000639).

last is "Men\u00e9ndez"
official_full is "Robert Menendez"
other_names['last'] is "Menendez"

It's odd that the last name has the accent, but the official full name does not. At his senate site (https://www.menendez.senate.gov/) and on wikipedia, he's shown without the accent. 

If he used to use the accent but now does not, last should be "Menendez" with other_names['last'] as "Men\u00e9ndez". 

Feel free to ignore this - I only noticed it because I was printing both last and official_full, and it seemed weird to me.

Also thanks for the awesome work!

JoshData:
Interesting. I'm pretty sure he used to use the accent / I half remember making that change (like 10 years ago). But more interesting is that he still uses it on the Spanish-language version of his official site: https://www.menendez.senate.gov/es even though it's not on the main site.

ezoehunt:
That is interesting. Seems like he's using the accent purposefully, so guess we should leave it alone.

JoshData:
I went back through the Wayback Machine and it doesn't appear he's ever used the accent on his main homepage (going all the way back to his House website in 2000). The accent doesn't appear on his name in any of my archives of bioguide data back to 2009. I also checked the [original source data](https://web.archive.org/web/20040411200054/http://opengov3.media.mit.edu:80/DBD/SRC/ROOTCSV/house.csv) for Members of Congress that I grabbed in 2003, and that was ASCII letters only. So I am baffled by when and why I would have added the accent. The accent was in the data when I first started version tracking my legislator data (in 2007).

So I'm OK to remove it.

PR #656, opened 2019-01-23
Remove accent from Robert Menendez's last name.

tcarobruce:
#492

For consistency with official full name, and other public sources.


JoshData:
:+1:

Issue #657, opened 2019-01-25
Religion data is very incomplete


YairRand:
The religion field is very incomplete. Pew has lists of religions of members of the 111th-116th congresses, mostly (entirely?) compiled from CQ Roll Call surveys:
* 111th: http://assets.pewresearch.org/wp-content/uploads/sites/11/2008/12/appendix.pdf
* 112th: http://assets.pewresearch.org/wp-content/uploads/sites/11/2011/01/relig-affil-pdf-updated-2-24-111.pdf
* 113th: http://assets.pewresearch.org/wp-content/uploads/sites/11/2012/11/113-congress-relig-affil.pdf
* 114th: http://www.pewforum.org/2015/01/05/members-of-congress-religious-affiliations/
* 115th: http://assets.pewresearch.org/wp-content/uploads/sites/11/2017/01/19161723/Member-affiliations-for-web.pdf
* 116th: http://www.pewforum.org/wp-content/uploads/sites/7/2019/01/Detailed-tables-NUMBER-CHECK-COMPLETE-1-3.pdf

Could this data be added?

JoshData:
I'm not sure if this would be right to add without permission from Pew and CQ.

dwillis:
I agree with Josh on this. Personally, I don't think it's something that we should be tracking with this repo, but I can understand the interest and potential research value.

JoshData:
If there's no objection, I will open a PR that deletes all of the religion data we have so there's no confusion.

dwillis:
No objection here.

JoshData:
Per our fairly limited consensus, we're not going to include religion data in this dataset at this time, so I'm going to close the issue. I'm sure we'd be glad to continue talking about a strategy to consolidate the data.

PR #659, opened 2019-02-03
remove the old religion data

JoshData:
This data came from my original import from GovTrack's legislator database, and the religion field probably came from my original import of data from the MIT Media Lab's Government Information Awareness project in 2003. The field was never maintained. To avoid confusion over an unmaintained field, per #657 and 9101feaae3ec45418d9e0806df656df7b74ff22e, this PR removes it.

### DIFF
--- a/executive.yaml
+++ b/executive.yaml
@@ -1500,7 +1500,6 @@
   bio:
     birthday: '1942-11-20'
     gender: M
-    religion: Roman Catholic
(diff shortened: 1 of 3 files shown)


dwillis:
Gonna give others a chance to weigh in on this, but I'm fully supportive.

JoshData:
:+1:

Issue #665, opened 2019-02-12
Possible incorrect or missing FEC ids


composerinteralia:
I found a couple of FEC ids that don't seem to match what I found on https://www.fec.gov:

* Joe Neguse
We have `H8CO02178`, which doesn't show up on https://www.fec.gov/data/candidates/?q=H8CO02178. Instead, I found https://www.fec.gov/data/candidates/?q=H8CO02160.

* David Trone
We have `H8MD06168`, which doesn't show up on https://www.fec.gov/data/candidates/?q=H8MD06168. Instead, I found https://www.fec.gov/data/candidates/?q=H6MD08549.

* Mitt Romney
We have `S4MA00143`, but I think that is an old one, since Romney is in Utah now I think it should be `S8UT00176` (https://www.fec.gov/data/candidates/?q=S8UT00176)

I will open a PR in case we want to make these changes.

PR #666, opened 2019-02-12
Add missing FEC ids

composerinteralia:
I found these ids while searching through
https://www.fec.gov/data/candidates/

For Joe Neguse and David Trone I couldn't actually find the FEC ids that
we currently have here on https://www.fec.gov/data/candidates/,
but I didn't remove them because I am assuming
they are coming from some other data source and it may make sense to do
some digging upstream to see why there is a mismatch.

For Mitt Romney we had the FEC id for Massachusetts,
but he is in Utah now.

Fixes https://github.com/unitedstates/congress-legislators/issues/665

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -38091,7 +38091,7 @@
 - id:
     bioguide: N000191
     fec:
-    - H8CO02178
+    - H8CO02160
     govtrack: 412761
     wikipedia: Joe Neguse
     wikidata: Q57242006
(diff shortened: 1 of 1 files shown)


dwillis:
Hey @composerinteralia, thanks! You can remove the current ones for Trone and Neguse; the ones you added are the correct ones. Can you update?

composerinteralia:
Updated!

PR #668, opened 2019-02-21
added 5 twitter accounts

amyewest:


dwillis:
@amyewest Thanks!

JoshData:
Thanks Amy! (Amy is GovTrack's research & communications manager.)

PR #671, opened 2019-03-04
Merge social media records for Bill Cassidy (C001075)

tcarobruce:
Easier for maintenance to keep all of a member's social media accounts under one record.


JoshData:
Thanks. cc @amyewest fyi.

PR #672, opened 2019-03-13
Update contact info

justinvasel:
This PR includes updated contact information for both House and Senate members as produced by the scripts `scripts/senate_contacts.py` and `scripts/house_contacts.py` respectively.

I had originally run the senate script because I noticed that Joni Ernst's office was outdated, and then I found that there were a handful of Senate offices that had changed, in addition to URLs and emails.

The House updates only include changes to members' full names.

I hand-audited roughly 50% of the office/URL changes to validate.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -37995,7 +37995,7 @@
     first: Gilbert
     middle: Ray
     last: Cisneros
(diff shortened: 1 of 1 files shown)


JoshData:
Great!

PR #674, opened 2019-03-22
Scrape committee membership

JoshData:
Although @dwillis did a lot of work on manually collecting committee membership, there is a lot of new data and some errors. We have a scraper, but now that the data has membership start dates and links to sources, the scraper needed to be updated to preserve existing data in the file.

I updated the scraper. The result was a lot of changes. So I ran it in stages (full committees first, then (missing) subcommittees we didn't have, then subcommittees we already had) so we have at least one commit with a readable diff.

I also added the House Select Committee on the Climate Crisis and the House Select Committee on the Modernization of Congress to the metadata since they are on the Clerk site and have members.

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -5450,7 +5450,7 @@ HSFA16:
   bioguide: H001085
   start_date: '2019-01-29'
   source: https://foreignaffairs.house.gov/press-releases?ID=1F0FB8B1-CCA6-4B3A-B360-22907C33D03B
-HSFA18:
+HSFA17:
 - name: Lee M. Zeldin
   party: minority
   rank: 1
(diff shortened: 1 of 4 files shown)


JoshData:
... I'm not sure how best to review this since one of the commits has a huge diff.

I think we also lost some _ex officiio_ entries which are indicated in source data @dwillis used but isn't in the clerk committee data.

dwillis:
I can live with losing the Ex Officio members, tbh.

dwillis:
@JoshData I'm in the process of reviewing these commits. Thanks for doing this!

dwillis:
Here's a fun one: SSCM26, which has different information about the minority members in the [Senate XML](https://www.senate.gov/general/committee_membership/committee_memberships_SSCM.xml) vs. the [committee site](https://www.commerce.senate.gov/public/index.cfm/pressreleases?ID=928E144C-B9C8-478A-92B7-29C5B270ECA5).

dwillis:
@JoshData I've done some clean-up on these files. Lemme know if that helps.

JoshData:
Thanks. I'll take a look. I'm traveling this week and most of next so I'm not sure if I'll be able to look soon tho.

dwillis:
@JoshData Any issues with this?

JoshData:
Sorry, lost track, all good. Merging.

PR #679, opened 2019-06-13
Add missing Open Secrets Candidate IDs

Haplo-Dragon:
I noticed that 19 of the current legislators were missing Open Secrets IDs, so I added them from the list of IDs available from Open Secrets [here](https://www.opensecrets.org/open-data/api-documentation). Please let me know if there's anything else I need to do, I'd be happy to help.


JoshData:
Thanks! Merging!

PR #681, opened 2019-07-09
Add missing Senate FEC ID's

MMcCormick:
Four current senators were missing Senate FEC ID's. Each of these had (correct but outdated) House of Represenatives FEC ID's, which may help explain why it went unnoticed by the scripts.


dwillis:
@MMcCormick thanks!

PR #682, opened 2019-07-24
Correct FEC ID for Susie Lee

MMcCormick:
I opted to change the existing ID, rather than add another, because I couldn't find H8NV03200 (the existing value) via search on fec.gov, and most representatives have just one ID per body of government. If it's preferable to keep the existing one as well, please let me know and I will make that change.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -39245,7 +39245,7 @@
 - id:
     bioguide: L000590
     fec:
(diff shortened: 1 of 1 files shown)


JoshData:
Maybe her ID was revised. LGTM but I'll let @dwillis approve since this is his area and he put in the ID originally (65e986c0). :)

PR #687, opened 2019-09-03
Update house websites script

tcarobruce:
The house_websites.py script was not finding any URLs due to a change in the HTML of the House directory page.

This change fixes the script, and updates current House member URLs based on the fixed script.

Most changes are just to remove a trailing slash or do http->https, but there are a few substantive changes as well.


JoshData:
Nice.

PR #691, opened 2019-09-04
Independent caucus optional

### LINKED ISSUE #689: caucus: ~  is invalid when party is Independent.
From circleci:

> legislators-current.yaml:A000367:terms[4](2019-01-03 to 2021-01-03): caucus: ~
 is invalid when party is Independent."

This refers to Justin Amash, who is indeed Independent and I believe did withdraw from the House Republican Conference. Maybe change the validation rule? Or add "caucus: None"? It is unusual, but there isn't actually some rule that Independents must caucus with somebody....

jeremydouglass:
Update tests and readme. Fixes #689


JoshData:
Thanks!

PR #692, opened 2019-09-18
add Rep. Murphy, Rep. Bishop

JoshData:
Neither one is on Bioguide yet. I guessed Murphy's bioguide ID from the next available M number; GPO bill XML is already giving the id for Biship. Bishop's birthday is from Wikipedia; Murphy's birthday is omitted --- I can't find a source.

address/office/phone is from running house_contacts.py.


csnardi:
https://www.congress.gov/help/field-values/member-bioguide-ids confirms Murphy's bioguide as M001210.

PR #698, opened 2019-10-05
icpsr ids - fix issue #601

### LINKED ISSUE #601: ICPSR IDs broken
Hi. I'm with UCLA Political Science and voteview.com -- we are your provider for ICPSR IDs. I noticed your `icpsr_ids.py` script a) hasn't been updated since the 113th Congress and b) uses out of date URL schemes. When UCLA took over voteview.com after Keith Poole retired from UGeorgia in 2017, we updated the site and thus the URL schemes.

In addition to changing the URLs, we are strongly encouraging people use CSV format files rather than the legacy ORD files we provide. We will be moving to formally deprecate the ORD files in the next year or so and they may be removed in the future.

My guess is you probably want this file for our data:
https://voteview.com/static/data/out/members/HSall_members.csv

If you would prefer individual congress files or individual chambers, the schema is `Cnnn_members.csv`, where `C` is the chamber(s) (H, S, or HS) and `nnn` is a zero-padded integer of length 3 for the number of the congress (i.e. 078 for the 78th congress, 004 for the 4th congress).

So, for the 110th House:
https://voteview.com/static/data/out/members/H110_members.csv

For the 54th Senate:
https://voteview.com/static/data/out/members/S054_members.csv

Relevant fields for mapping to your data:
`bioname`, `bioguide_id`, `chamber`, `icpsr`, `state_icpsr`, `state_abbrev`.

Detailed metadata for the fields provided:
https://voteview.com/articles/data_help_members

Hope this helps. I may have some time to write a PR to replace your script with a current one but in case anyone is actively working on this project I figure I'd open the issue for now.

asebold:
Fixes #601
Current urls for getting Vote View data are being depreciated. This PR updates the urls in icpsr_ids.py script to adhere to new format. Script also had to be updated to handle the new data being returned by Vote View. Ran the script and updated icpsr ids for legislators-current.yaml. Now, the icpsr_ids.py script correctly queries Vote View and sets the icpsr id for current and historical legislators (based off congress number passed to script).

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -36933,6 +36933,7 @@
     maplight: 2246
     google_entity_id: kg:/m/0gfgsrm
     opensecrets: N00040712
+    icpsr: 21750
   name:
     first: Ron
     last: Estes
@@ -36976,6 +36977,7 @@
     google_entity_id: kg:/m/030pb5p
     house_history: 15032448990
     maplight: 2247
+    icpsr: 21751
   name:
     first: Greg
     last: Gianforte
@@ -37019,6 +37021,7 @@
     google_entity_id: kg:/m/0f9kbx
     house_history: 15032448994
     maplight: 2248
+    icpsr: 21753
   name:
     first: Ralph
     last: Norman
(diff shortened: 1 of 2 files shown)


JoshData:
Thank you!!

PR #704, opened 2019-12-06
Bugfix Term Dates for Frank J. Cannon

nmacri:
Updates House and Senate term dates to reflect body text in bioguide

**Citation**

> elected as a Republican to the Fifty-fourth Congress and served from March 4, 1895, to January 4, 1896, when the Territory was admitted as a State into the Union; was then elected as a Republican to the United States Senate and served from January 22, 1896, to March 3, 1899
http://bioguide.congress.gov/scripts/biodisplay.pl?index=C000118


JoshData:
The "1895-12-02" start date probably should remain unchanged. It reflects the start date of the first session, which is our convention for start dates.

nmacri:
Okay, cool.  Updated!

JoshData:
Thanks!

PR #709, opened 2020-01-07
Rep. Duncan Hunter resigned - effective Jan 13

JoshData:
https://twitter.com/ChadPergram/status/1214671293213290496


dwillis:
Do we change his `end_date` for folks who resign (even though he's being removed from the file)?

JoshData:
Prior to the date of resignation? I don't think we've done that in the past out of laziness, but it would be fine by me. (Although it will create a conflict with this PR....)

dwillis:
Yeah, that's what I meant. If we don't do it, that's fine.

PR #713, opened 2020-01-17
Sen. Loeffler Social Media

putorti:
Her Instagram, Youtube accounts could be campaign/personal; but Facebook and Twitter seem to be office accounts.


legislators-social-media.yaml
```diff
@@ -4936,3 +4936,9 @@
     youtube_id: UCVroMJWH3jsh4shPSlp24sw
     twitter_id: 346509049
     instagram: michael.fq.san.nicolas
+- id:
+    bioguide: L000594
+    govtrack: 456790
```
csnardi: I think you need the `social:` section here (and then indent the following lines).

PR #714, opened 2020-02-12
Add missing IDs for Twitter handles

sdmccabe:
This PR adds missing Twitter IDs only for legislators who already have
a Twitter handle present in the data. I don't know if these are campaign
accounts or not; these are just the IDs for accounts that are already
there.


JoshData:
Thank you merging!

PR #720, opened 2020-05-09
Historical fixes

JoshData:
I was looking at whether Rep. Mfume's 24-year absence from Congress was a record (no, Maryland congressman [Rep. Phillip Thomas](https://www.govtrack.us/congress/members/phillip_thomas/410763) seems to hold that record, with a 34-year absence between 1841-1875).

In the process I found some data errors that all appear to be duplicates of the entire terms of representatives hidden in similar-named representatives. These edits seem to date back to my original 2009 parse of Bioguide. This PR removes those duplicates and makes some fixes that I found while researching the errors.

### DIFF
--- a/test/validate.py
+++ b/test/validate.py
@@ -386,7 +386,7 @@ def check_id_uniqueness(seen_ids):
 def check_district_offices():
     has_errors = validate_offices(skip_warnings=True)
     if has_errors:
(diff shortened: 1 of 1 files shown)


dwillis:
Talk about going down your rabbit holes! Nice work.

Issue #185, opened 2014-04-17
Add Congress Number to legislators YAML Files "Terms"


dmil:
Currently a term for a legislator in the YAML file term currently looks like this :

```
- type: rep
  start: '1993-01-05'
  end: '1994-12-01'
  state: OH
  district: 13
  party: Democrat
```

I want to add a field called "congress" (in this case the value would be 103).

My question is, how can I do this in a way that will allow me to contribute it back to the repo. I figured a yaml file to map date ranges to congress numbers and then a script to update every record in the existing yaml files is a start, but if you can think of a better way, I can do that.

Unipartisan:
Josh, you say, "it's actually not possible to assign a congress to every term."   I assume that is because a term is longer than a congress for senators.  What I would find very useful would be to group currently serving members of congress by the congress number rather than to label a historical file with a congress field.

As it currently stands, we have legislators-current.yaml and legislators-historical.yaml.  If I wanted a copy of the legislators-current.yaml pertaining to 'current' members during the 113th congress, I can only get that by browsing back through the git repository to an earlier state because the current one is for the current congress, 114.

What if, at the end of a given congress, the legislators-current.yaml file were finalized and saved as legislators-113.yaml, legislators-114.yaml, etc. moving forward.

If a term were broken mid-congress my recommendation would be to simply add the successor into the list so that anyone who served any time during that congress would be listed.  Doing this would allow us to classify legislators by congresses without splitting up date ranges for the terms.

A major problem I have come across in my work in regards to the legislators-historical.yaml file is a lack of a certain id field 'lis'. For example "lis: S136" which only appears for senators.  This 'lis' field is in the legislators-current.yaml, but is not in the legislators-historical.yaml.

Here is why that 'lis' field matters so much for someone who may be trying to do what I am.

If you look at this .json file for sequential vote number 281 in the senate in 2013

https://www.govtrack.us/data/congress/113/votes/2013/s281/data.json

You see senators listed under Nay and Yea in the following form:

        "display_name": "Alexander (R-TN)", 
        "first_name": "Lamar", 
        "id": "S289",   (<---this is the equivalent of 'lis' in legislators-current.yaml)
        "last_name": "Alexander", 
        "party": "R", 
        "state": "TN"

The 'lis' field from legislators-current.yaml, which does not appear in legislators-historical.yaml is the only id type that links the voter to their respective listing in the legislator data. ('lis' is listed as "id" above)

[...]

PR #721, opened 2020-05-10
Fix term date errors that prevented assigning Congress numbers to terms

JoshData:
...

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -1824,7 +1824,7 @@
     district: 1
     party: Republican
   - type: sen
-    start: '1994-01-01'
+    start: '1994-11-17'
     end: '1997-01-03'
     state: OK
     class: 2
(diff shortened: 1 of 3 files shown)


JoshData:
Ok so one of our old issues (#185) was that we wanted to put a Congress number (or range) on each term. But the historical data had a lot of errors (mostly from senate appointments, possibly because my original scrape of bioguide didn't work right) where the date ranges on the terms didn't line up with valid term ranges, like senate terms longer than six years, or missing term boundaries where we know there should be according to the senate class.

This PR adds a test that checks that each term corresponds to a Congress number or valid range and fixes all of the remaining data errors so that now the test passes.

After this, we can go back and solve #185.

JoshData:
I added a second commit that fixes overlapping terms in historical data. We had a test that was disabled for historical data because of these errors. Now the test is enabled and there are no failures.

PR #724, opened 2020-05-20
Update CircleCI docker image to Python 3.8, drop -browsers

JoshData:
There was a probably-transient error in #723. This probably has nothing to do with that, but is probably a good idea anyway.


JoshData:
The build step passed so I'll merge right away and hope that the deploy step works.

PR #728, opened 2020-06-09
Fix Millard Fillmore's party for 1833 House Term to be Whig, not anti-Masonic

yahelc:
Millard Fillmore was elected in 1832 as anti-Masonic, but before taking office had formed the Whig Party.

https://history.house.gov/People/Detail/13119

By the standards of the data dictionary:
>If the legislator changed parties, this is the most recent party held during the term and party_affiliations will be set

Given it seems like he never served in Congress as anti-Masonic and changed during this term, it seems like the data should have him as Whig for this term.


JoshData:
Thanks!

PR #730, opened 2020-06-18
legislators-historical.yaml: Add missing Congresspeople

dshearer:
This was scraped from congress.gov/members. It just gives the term
years, so I just put '01-03' for the month/day.


JoshData:
There are no missing congresspeople in our data. The legislators you added are currently serving and so are in the legislators-current.yaml file.

dshearer:
Ooops.

PR #733, opened 2020-07-21
Adds Chris Jacobs

dwillis:
Doesn't have govtrack id (cc @JoshData)


JoshData:
Added, merging, thanks!

PR #734, opened 2020-09-08
Remove ; from address after office_building_full

jrdnbradford:
Creates consistent address format in data. House contact address data currently includes ; after the office building whereas Senate contact address data does not.


JoshData:
Could you also run the updated script and commit the changes?

jrdnbradford:
Of course! I'll close this PR. I've opened PR #735 that commits both changes.

PR #740, opened 2020-11-10
Kelly sworn in replacing McSally; bump Loeffler end date to the run-off election date; Hall

JoshData:
When I added Loeffler in ea7c90d7, I set her end date to the special election date, since usually replacement senators take office immediately after that. But that has come and gone and she is still the appointed senator.


JoshData:
Continued in the PR with adding Mark Kelly and McSally removed.

JoshData:
LIS ID added.

PR #751, opened 2021-01-12
Conform names to how they appear in House.gov and Senate.gov XML, mostly

JoshData:
I'm open to discussion here. Some of the changes here appear to correct clear errors or very confusing situations where legislators may not be recognizable by their legal name. Other changes are hard to say if it's necessarily better or not, depending on how formal we are trying to be.

* Where the House/Senate is more informal than the legal name in our data, the informal first name is put into the nickname.
* Where the legislator appears to not go by their legal first name, their legal first name is replaced with their name according to House.gov/Senate.gov, in most cases.
* Some middle initials are expanded to full middle names and some missing middle names/initials are added from the House.gov/Senate.gov data.
* Since the House/Senate data appears informal, where the data is more formal than what we have, the formal name is chosen and a nickname is dropped.
* Suffixes that are not used are removed.


JoshData:
Since/if there are no objections, I'll merge this tomorrow along with #754 at noon.

PR #752, opened 2021-01-16
Rep. Richmond resigned yesterday to join the Biden Administration

JoshData:
Additionally fixing committee-membership-current.yaml which should have been made an empty mapping and not a list in ee290b5dcc1696bc575e31212501766ad7fd7fde which cleared it out for the new Congress.


JoshData:
Merged via #753 which included these commits.

PR #753, opened 2021-01-17
Harris to resign tomorrow

JoshData:
https://apnews.com/article/election-2020-biden-inauguration-joe-biden-biden-cabinet-hispanics-39f417251dd1de8d3a8e178f47f186b2


JoshData:
The resignation was effective at noon, so merging.

PR #754, opened 2021-01-17
Add executive.yaml terms for Biden and Harris

JoshData:


JoshData:
If 10 minutes early is good enough for the inauguration it's good enough for me...

PR #755, opened 2021-01-18
Add nickname for Barry Moore

nickoneill:
AL 2's Felix Barry Moore goes by Barry Moore. While this is already reflected in `official_full`, for other legislators this field is very formal. We've been using the nickname + last to construct a more casual name, I think this change fits with the convention in the rest of the file.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -40176,6 +40176,7 @@
     first: Felix
     middle: Barry
     last: Moore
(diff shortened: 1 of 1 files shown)


JoshData:
I've made that change in #751.

nickoneill:
Makes sense! Thanks for the detail for how you think through this, it makes sense in the cases that I know about.

PR #757, opened 2021-01-20
Senators Padilla filling Harris's recent vacancy, Warnock replacing Loeffler whose term now ends, and Ossoff

JoshData:
If bioguide IDs become available before the end of the day I'll add them and remove the commit that removes the bioguide validation.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -39428,40 +39428,6 @@
     office: 313 Cannon House Office Building
     phone: 202-225-3415
 - id:
-    bioguide: L000594
-    lis: S405
-    fec:
-    - S0GA00526
-    govtrack: 456790
-    wikipedia: Kelly Loeffler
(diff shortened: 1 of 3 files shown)


jlev:
FYI, new senator phone numbers are:

Padilla: 202-224-3553
Ossoff: 202-224-3521
Warnock: 202-224-3643

per https://www.senate.gov/general/contact_information/senators_cfm.cfm

JoshData:
Apologies - I'm feeling a little sick so I may not be able to finish this PR tonight. If someone else wants to grab the bioguide IDs from bioguide (they're up now) and then dump the last revert commit off of this PR, it'll be good to go. The Senate will hold a vote tonight I think so the LIS IDs will also become available then from the Senate vote data.

JoshData:
... And the phone numbers can probably be updated by running our senate_contacts.py script.

PR #760, opened 2021-01-24
Fix last and middle name for Cathy McMorris Rodgers

asebold:


JoshData:
That's how we used to have it, but it is not how she appears on House.gov anymore:

image

image

The switch from "McMorris Rodgers" to "Rodgers" as her last name appears to have happened in 2019 between roll calls 15 and 16.

asebold:
Ugh, annoying. I reached out to the house.gov webmaster about it.

GPHemsley:
@asebold Any luck?

(For the record, the screenshot above is of [https://www.house.gov/representatives#name-r](https://www.house.gov/representatives#name-r).)

asebold:
@GPHemsley I remember reaching out to them but I don't believe I ever got a response.

GPHemsley:
@asebold Well it looks to still be the case, so I doubt this was a mistake. Want to convert this to an `other_names` sections instead?

PR #767, opened 2021-02-11
Add Rep. Tenney

JoshData:


JoshData:
Ooof. The tests showed I messed this up, not realizing she has previously served in Congress...

PR #769, opened 2021-02-15
add marjorie taylor greene. 

cbenin0711:
hi! wrong tag i wasnt sure how to do this, not someone who codes. but please add her to the list (I am curious about other names vs legal) thank you1


dwillis:
Hi @cbenin0711 - thank you for the contribution, but we already have Marjorie Taylor Greene in the data files.

PR #770, opened 2021-03-02
Added House standing full committee assignments

dwillis:
This is based off [this Clerk document](https://clerk.house.gov/committee_info/scsoal.pdf) plus several subsequent House resolutions naming members. It does not include Marjorie Taylor Greene, who was removed from two committees, or Ron Wright, who died in early February. It also is currently missing the House select committees (Intel, Climate and Modernization) and House assignments to joint committees.

If folks want to wait until we get those to merge, that's ok with me.

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -2853,3 +2853,3390 @@ JCSE:
   rank: 4
   bioguide: T000476
   chamber: senate
+HSAG:
+- name: David Scott
+  party: majority
+  rank: 1
+  title: Chairman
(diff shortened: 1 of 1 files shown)


JoshData:
I ran lint.py to hopefully get it to pass tests. It should only have re-ordered the committees into sorted key order, I think. Look ok? If tests pass now I'm fine to merge.

dwillis:
@JoshData looks good to me.

JoshData:
Great. And thanks!!

PR #781, opened 2021-05-12
Update Senate committee membership (House committee data remains unav…

JoshData:
…ailable)


JoshData:
The diff appears very large here but if you run `git show --minimal` it's actually not that many changes.

PR #785, opened 2021-06-16
fix circleci config to run deploy step again

jamesturk:
it looks like the branch filter was left as master despite the name changing to main


JoshData:
Ahha thanks!

PR #786, opened 2021-06-16
a few other fixes where master should have been changed to main

jamesturk:
it looks like the publish step is still failing, updated master->main in that script as well as a few others


jamesturk:
ah you got to it first 👍

JoshData:
But not the rest. Still merging. :)

PR #788, opened 2021-07-19
YouTube channel IDs for House and Senate committees, where applicable

ericisaiah:
I found these useful for a project and thought it would save others time if they were incorporated into the base repo.

Although legislators' social media accounts are in a separate yaml file, I suggest putting the committees' YouTube channel ID directly on the current committees list both for convenience and since it does not need to be updated with the same frequency as the legislators list.

Also worth noting that Senate committees largely do not have YouTube channels, but in cases where I could find those I added them, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ Additional fields are present on current committee entries (that is, in `committ
 * phone: The phone number of the committee.
 * rss_url: The URL for the committee's RSS feed.
 * minority_rss_url: The URL for the committee's minority party website's RSS feed.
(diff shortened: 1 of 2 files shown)


JoshData:
Would you also update README.md to document that this new field exists?

ericisaiah:
Sure thing. I just added a line to the committee data dictionary, but let me know if you had something else in mind.

JoshData:
Great, thank you!

ericisaiah:
No problem. Thanks for maintaining this great resource.

PR #789, opened 2021-07-20
Add OpenSecrets ids to about 65 legislators; correct FEC id for Sara …

admarwick:
…Jacobs

-- Roughly 65 legislators had no -id/opensecrets entry; added them using data from the OpenSecrets website (CRP_IDs.xls, downloaded from https://www.opensecrets.org/resources/create/data_doc.php).
-- Corrected the fec id (-id/-fec) for Sara Jacobs; the previous id is not listed by the fec.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -39641,6 +39641,7 @@
     lis: S412
     fec:
     - S0AL00230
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!

PR #797, opened 2021-11-08
Fix a typo and add a missing progress message in formatting script

clarkep:
In the alternate bulk format script, we were printing that we were converting the files to YAML when we are really converting them to CSV. Also, one of the messages that said this was missing.


clarkep:
I hope this isn't too trivial

JoshData:
Thank you.

Issue #807, opened 2022-01-25
Rep. Gregory Murphy has resigned from committees


syntheticgoo:
Rep. Gregory F. Murphy has resigned from Education and Veterans committees today.

PR #808, opened 2022-01-25
Gregory F. Murphy has resigned from committees

syntheticgoo:
Ticket #807 .

This can be verified in the congressional record once posted for today: https://www.congress.gov/congressional-record/2022/01/25/daily-digest

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -6933,10 +6933,6 @@ HSED:
   party: majority
   rank: 12
   bioguide: M001206
-- name: Gregory F. Murphy
-  party: minority
-  rank: 12
-  bioguide: M001210
 - name: Susan Wild
   party: majority
   rank: 13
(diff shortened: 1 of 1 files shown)


JoshData:
We use a script to maintain the committee membership files so this would be lost if we ran the script, which we should do since there are other updates. The easiest thing to do is wait for the change to appear in the clerk.house.gov data.

JoshData:
It looks like he was moved to Ways and Means: https://www.govtrack.us/congress/bills/117/hres875/text

If you can add that to the PR, I'll merge it. Seems like the clerk.house.gov committee membership info is actually just out of date.

GPHemsley:
It looks like running `committee_membership.py` will pick this change up now (along with a number of others), though it would revert #820 because Rep. Young's death hasn't been accounted for yet.

JoshData:
Thanks. I ran the script and then sweep.py to remove legislators no longer serving (and pushed to main).

PR #816, opened 2022-02-20
Update congress people twitter handles

fre5:


JoshData:
I removed one --- we don't include campaign accounts in this repository. Thanks for the rest!

PR #817, opened 2022-02-24
Fixed family members with " or ' in the name

simonw:
These two records display `?` where they should have `"` or `'`.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -17500,7 +17500,7 @@
     office: 1236 Longworth House Office Building
     phone: 202-225-4965
   family:
-  - name: "Thomas D'Alesandro Jr."
+  - name: Thomas D'Alesandro Jr.
     relation: daughter
 - id:
     bioguide: P000593
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -17500,7 +17500,7 @@
     office: 1236 Longworth House Office Building
     phone: 202-225-4965
   family:
-  - name: Thomas D?Alesandro Jr.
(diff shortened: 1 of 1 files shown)


simonw:
Here's the test failure:
```
python test/are_files_linted.py

legislators-current.yaml needs to be linted:
--- in repository
+++ after linting
@@ -17500,7 +17500,7 @@
     office: 1236 Longworth House Office Building
     phone: 202-225-4965
   family:
-  - name: "Thomas D'Alesandro Jr."
+  - name: Thomas D'Alesandro Jr.
     relation: daughter
 - id:
     bioguide: P000593
legislators-historical.yaml needs to be linted:
--- in repository
+++ after linting
@@ -442757,7 +442757,7 @@
     office: 2079 Rayburn House Office Building
     fax: 202-225-2486
   family:
-  - name: "E. G. \"Bud\" Shuster"
+  - name: E. G. "Bud" Shuster
     relation: son
 - id:
     bioguide: S000583

Exited with code exit status 1
```
I disagree with the linter here - I think it should allow fields like this to be quoted.

simonw:
Looks like https://github.com/unitedstates/rtyaml would prefer it to look like this:
```pycon
>>> import rtyaml
>>> print(rtyaml.dump({"family": [{"name": 'E. G. "Bud" Shuster'}]}))
family:
- name: E. G. "Bud" Shuster
```
I'll try that instead.

simonw:
OK, `rtyaml` is better at YAML than I am! Turns out you don't need to wrap strings like that in double quotes after all.

JoshData:
Thanks!

PR #819, opened 2022-02-27
Merge two Joe Biden records in executive.yaml

### LINKED ISSUE #818: Duplicate Joseph Biden records in executive.yaml
Joe Biden is listed twice in that file. First, for his two terms as vice president:

https://github.com/unitedstates/congress-legislators/blob/ffc37dd964dbbdba9d7855375d62ae712a679341/executive.yaml#L1484-L1499

And secondly for his term as president:

https://github.com/unitedstates/congress-legislators/blob/ffc37dd964dbbdba9d7855375d62ae712a679341/executive.yaml#L1584-L1603

Note that these records are also inconsistent with each other - for example one lists his middle name as `Robinette` and the other as `R.`.

simonw:
Fixes #818

### DIFF
--- a/executive.yaml
+++ b/executive.yaml
@@ -1491,6 +1491,11 @@
     icpsr: 14101
     fec:
     - S8DE00012
+    cspan: 34
+    wikipedia: Joe Biden
+    wikidata: Q6279
+    google_entity_id: kg:/m/012gx2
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!

PR #821, opened 2022-03-14
Update build status badge in README for CircleCI

jamesa:
It seems like the Travis-CI integration is no longer used, so this change updates the badge shown in the README to show the CircleCI badge instead.


JoshData:
Thanks!

PR #829, opened 2022-04-08
Revise Rep. Cawthorn's first name to Madison

JoshData:
Per 1c13038c407069393bd578009cb025700e57da2c, our policy is to make the first name field of legislators a name that would be recognized by the general public as the legislator's first name, even if it is not the legislator's legal first name, especially to follow the legislator's preferred first name in the Bioguide or House/Senate websites (which show "Madison" in this case).

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -40136,8 +40136,7 @@
     google_entity_id: kg:/g/11jcrrfg9t
     ballotpedia: Madison Cawthorn
   name:
-    first: David
-    nickname: Madison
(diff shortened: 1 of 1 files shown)


GPHemsley:
It would be good if there was a place to put the legal name so that it's at least available.

JoshData:
i happened to read this recently which talks about how there really isn't such a thing as a (unique) legal name:

http://hrlr.law.columbia.edu/hrlr/there-is-no-such-thing-as-a-legal-name/

PR #834, opened 2022-05-15
Updated Rubio district offices

nickoneill:
Tampa changed locations + phone, cleaned up some other unlisted fax numbers for the rest of the locations.

From: https://www.rubio.senate.gov/public/index.cfm/contact


nickoneill:
Thanks!

PR #835, opened 2022-05-22
Add Open Secrets identifier for Jake Ellzey

poliquin:
Added an identifier for Jake Ellzey based on [his Open Secrets page](https://www.opensecrets.org/members-of-congress/jake-ellzey/summary?cid=N00042243&cycle=2022).


JoshData:
Thanks!

PR #842, opened 2022-08-03
Rep. Walorski was killed today

JoshData:
https://www.politico.com/news/2022/08/03/indiana-gop-rep-walorski-three-others-die-in-auto-accident-00049665

(It feels a little weird to open the PR so quickly, but I wanted to get it done while I had a moment.)


dwillis:
Awful news.

JoshData:
Yeah.

PR #847, opened 2022-09-14
Add Rep. Mary Peltola

putorti:
Sworn in on September 13, 2022

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -40919,3 +40919,26 @@
     district: 1
     party: Republican
     phone: 202-225-2472
+- id:
+    bioguide: F000476
+    fec:
+    - H2AK01158
+    govtrack: 456870
+    wikipedia: Mary Peltola
+    wikidata: Q13562231
+    ballotpedia: Mary Peltola
+  name:
+    first: Mary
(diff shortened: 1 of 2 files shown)


legislators-current.yaml
```diff
@@ -40919,3 +40919,26 @@
     district: 1
     party: Republican
     phone: 202-225-2472
+- id:
+    bioguide: F000475
```
JoshData, line 40923: That's not right. :)

putorti: Sorry about that, good now?

dwillis: @putorti The `fecid` is the candidate ID, not the committee ID. Can you fix?

putorti: @dwillis Sorry, my first new official; fewer mistakes next time!

JoshData:
Hijacking your branch a bit :), I made some additional corrections and also added the two other reps sworn in last week. @dwillis Would you look it over?

dwillis:
Looks good to me - good idea to add a validator for FEC IDs, @JoshData .

dwillis (approved):

PR #851, opened 2022-10-25
fix: Correct contact form URL for Senator John Hoeven and Tina Smith

MiniCodeMonkey:
The previous URL was 404'ing

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -10495,7 +10495,7 @@
     url: https://www.hoeven.senate.gov
     address: 338 Russell Senate Office Building Washington DC 20510
     phone: 202-224-2551
-    contact_form: http://www.hoeven.senate.gov/public/index.cfm/email-the-senator
+    contact_form: https://www.hoeven.senate.gov/contact/email-the-senator/form
     office: 338 Russell Senate Office Building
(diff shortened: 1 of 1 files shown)


legislators-current.yaml
```diff
@@ -10495,7 +10495,7 @@
     url: https://www.hoeven.senate.gov
     address: 338 Russell Senate Office Building Washington DC 20510
     phone: 202-224-2551
-    contact_form: https://www.hoeven.senate.gov/contact/email-the-senator/form
+    contact_form: http://www.hoeven.senate.gov/public/index.cfm/email-the-senator
```
JoshData, line 10498: This is for a previous term so unless you know that the URL we had was incorrect at the time (i.e. January 2017), this should be put back.

JoshData (changes requested):

legislators-current.yaml
MiniCodeMonkey: @JoshData Thanks for reviewing! I believe this was corrected in my second commit. The first commit did indeed change the `contact_form` for a previous term.

MiniCodeMonkey: No worries!

PR #822, opened 2022-03-19
Update Twitter accounts

GPHemsley:
Add account for Rep. David G. Valadao.

Update accounts for Sen. Rand Paul and Sen. James M. Inhofe.

Add 'home' and 'intent' to Twitter blocklist.

----

Justification:
* [\@RepDavidValadao](https://twitter.com/RepDavidValadao) links to Valadao's House website, which links back.
* [\@SenRandPaul](https://twitter.com/SenRandPaul) identifies itself as an official government account and links to Paul's Senate website, which links back. [\@RandPaul](https://twitter.com/RandPaul), on the other hand, links to Paul's campaign website.
* [\@JimInhofe](https://twitter.com/JimInhofe) links to Inhofe's Senate website, which links back. [\@InhofePress](https://twitter.com/InhofePress) also links to it. His campaign account appears to be [\@inhofeforsenate](https://twitter.com/inhofeforsenate).

Not included in this update:
* Sen. Mitch McConnell maintains two different accounts: [\@LeaderMcConnell](https://twitter.com/LeaderMcConnell) for his role as Republican leader and [\@McConnellPress](https://twitter.com/McConnellPress) for his role as the senator from KY. Each links to its corresponding Senate website, which links back.
* Rep. Jamaal Bowman appears to be conflating his official and campaign accounts. [\@RepBowman](https://twitter.com/RepBowman) appears to be his official account, which links to his House website (via linktree), but his House website links to [\@JamaalBowmanNY](https://twitter.com/JamaalBowmanNY), which in turn links to his campaign website (via myurls).
* It is not clear whether Sen. Roger Marshall has an official account. [\@SenatorMarshall](https://twitter.com/SenatorMarshall) has been stripped of all information and [\@RogerMarshallMD](https://twitter.com/RogerMarshallMD) does not link to any websites. However, Marshall's Senate website does link to the latter.

PR #858, opened 2022-12-09
Add Tim Kaine twitter

mrclay:
Linked from https://www.kaine.senate.gov/news


GPHemsley:
Per Past @JoshData in #822, https://tweeterid.com/ is the resource, and it reports `1561739334230892544`.

eliotpearson:
https://www.youtube.com/watch?v=KS6f1MKpLGM

Just saying...

JoshData:
The value is no longer correct.

PR #862, opened 2022-12-25
2022 Election Update

JoshData:
From me and the rest of the GovTrack team (@amyewest and Jesse).

Marking as a draft so we don't merge it before Jan 3 (~ at noon).


JoshData:
The main thing missing is bioguide IDs which aren't published yet as far as I'm aware.

dwillis:
@JoshData Senate XML has bioguide IDs for new folks: https://www.senate.gov/general/contact_information/senators_cfm.xml

JoshData:
Ah thanks! I can complete it in a few hours. (If anyone has a more urgent need to complete it and wants to push a commit, I don't mind.)

dwillis:
@JoshData have added those four.

JoshData:
There's unfortunately a little hiccup with the bioguide ID for Rep. Donald Davis: D000230 duplicates the ID for a historical legislator. The actual Bioguide no longer has an entry for D000230 because the entry was merged into https://bioguide.congress.gov/search/bio/L000365 because they are the same person (according to the entry). I emailed the Clerk's office to double check this was intentional.

JoshData:
It was intentional, so I removed the historical entry for the rep for bioguide ID D000230 and gave it to the new rep.

Merging.

PR #871, opened 2023-02-01
Senator Pete Ricketts

putorti:
Ben Sasse [resigned on January 8, 2023](https://bioguide.congress.gov/search/bio/S001197) and Pete Ricketts was appointed to fill his seat on [January 23, 2023](https://www.ricketts.senate.gov).

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -32538,55 +32538,43 @@
     office: 502 Hart Senate Office Building
     phone: 202-224-5842
 - id:
-    bioguide: S001197
-    thomas: '02289'
+    bioguide: R000618
+    thomas: ''
     fec:
-    - S4NE00090
-    govtrack: 412671
-    opensecrets: N00035544
-    lis: S382
-    maplight: 2127
-    wikidata: Q16192221
-    wikipedia: Ben Sasse
-    google_entity_id: kg:/m/0j5wk_6
-    votesmart: 150182
-    cspan: 77429
-    icpsr: 41503
-    ballotpedia: Ben Sasse
+    - S6NE00129
+    govtrack: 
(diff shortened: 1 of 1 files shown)


putorti:
@JoshData or @dwillis, I realize I'm missing some of the fields; I did the best I could to locate them, but, I was unsure how to get the LIS key from the Senate roll call page; there's no GovTrack ID yet; etc. I'm hoping someone can take this over the line, and hopefully this is at least better than filing an issue.

JoshData:
Thanks for getting this started, Jason.

Your checked-out files from this repo were out of date --- I had already zapped Sasse --- making it difficult to complete the PR and get a clean merge. So I started anew in #873 but I copied the IDs you found.

For reference, and others who may be reading-

* The `thomas` ID from the old thomas.loc.gov site is no long a thing, so it can be skipped.
* Anyone is free to mint a new GovTrack ID by just assigning the next available number (which you can usually guess by looking at the previously last entry and adding one, the tests that run on GitHub should check that it's not already in use).
* The `lis` ID can be found in the XML link on Senate vote pages. In this case at https://www.senate.gov/legislative/LIS/roll_call_votes/vote1181/vote_118_1_00005.xml:

image

* The first name should be the name the legislator goes by if they use it e.g. on their official website otherwise people get confused.
* In cases like this I like to set the end date to the date of the special election in which their term ends. It's a little hazy exactly when their term ends since the results may not be known on that date, but it can be adjusted when it's known. We also have `how: appointment` and `end-type: special-election` which GovTrack recognizes to alter how it describes how long the person is serving (this is noted in `misc/new-member-template.yaml`).

putorti:
@JoshData Thanks Josh, much appreciated!

PR #885, opened 2023-03-30
edit senator tillis first name

joevus:
It looks to me like Sen Tillis's first name is Thomas per https://bioguide.congress.gov/search/bio/T000476

Noticed this while joining a custom dataset with a dataset from this repo. This repo is great by the way, thanks.


dwillis:
@joevus thanks - we try to reflect the names that the senators use, and in this case Tillis has long gone by "Thom", including on his official website: https://www.tillis.senate.gov/.

PR #887, opened 2023-05-13
Correct typo in McConnell's current leadership

mattpaz:
This PR corrects what might have been a prior typo made when updating current leadership roles


JoshData:
Thanks!

Issue #886, opened 2023-05-13
Missing Committee Members


mattpaz:
There seem to be many subcommittees (like `HSSY20`) that don't include the members.  
If these are manually curated, I'd image a PR would be welcome?

https://github.com/unitedstates/congress-legislators/blob/main/committee-membership-current.yaml#L4476

JoshData:
It's updated by https://github.com/unitedstates/congress-legislators/blob/main/scripts/committee_membership.py from XML data from the House and Senate. We run the script manually periodically. I just ran it and pushed the update.

mattpaz:
Ah, cool. I didn't realize that xml data existed.  Thanks for pointing me in the direction of the python code and for running an interim update too.

The updates definitely filled in some gaps!

I'll take a deeper look at the python script and upstream sources at some point.
In the meantime, I'll just cite some of those that still have no committee members listed.

```
HLIG03
HLIG05
HLIG08
HLIG10
HSBA13
HSBA15
HSED07
HSGO02
HSGO25
HSGO28
HSGO29
HSHA01
HSII22
HSRU05
HSSM25
HSSM26
SSCM26
SSCM27
SSCM28
SSCM29
SSCM30
SSCM31
SSCM32
SSEV16
SSEV18
SSFR12
SSFR13
SSGA15
SSGA16
SSGA17
SSGA18
SSGA19
```

Looking at the `committees_current` source, it seems that maybe it simply needs to be reviewed/updated before re-running the script?

I wonder if we could use this: 
https://api.congress.gov/#/committee/committee_list
as a good source for updates?

I'll try to compare the data in the congress.gov apis to the data in committee_current at some point (if someone doesn't beat me to it.

---

Thanks again!

JoshData:
It's also possible the subcommittee has no members or that it exists only on paper. But defunct might be the most likely.

mattpaz:
I've completed a review of the defunct committees and have a PR started here:
https://github.com/unitedstates/congress-legislators/pull/890

Take a look and let me know if I missed anything.

mattpaz:
With https://github.com/unitedstates/congress-legislators/pull/890 committed, I think we can probably close this issue.

PR #889, opened 2023-05-21
Run 2023-05-21 updates from wikidata_update script

mattpaz:
This update contains:
*  updates from the `wikidata_update.py` script. 
*  additional error handling for the `wikidata_update.py` script to better handle upstream errors in votesmart data entries sourced from wikidata.

---

Incidentally, I observed that there are 81 records in `legislators-current.yaml` that don't have an `opensecrets_id`. At first glance, they all seem to be new entrants from the 118th congress. If there isn't a special mechanism for updating those ids, I might look at sporadically populating them on wikidata and creating a subsequent PR after each incremental update.

Secondarily, I wonder if a github action that would run `wikidata_update.py` on cron and create an automated PR might be interesting?


JoshData:
Thanks!

The two JSON files in the alternate_formats directory should be removed. They don't exist on the main branch. They're generated by CI on the gh-pages branch only.

> 81 records in legislators-current.yaml that don't have an opensecrets_id. At first glance, they all seem to be new entrants from the 118th congress. If there isn't a special mechanism for updating those ids, I might look at sporadically populating them on wikidata and creating a subsequent PR after each incremental update.

Yeah that would be great. Sometimes legislators don't have opensecrets IDs when they are sworn in, usually after special elections. And in other cases I may have just been too busy to look it up when adding the legislator.

> I wonder if a github action that would run wikidata_update.py on cron and create an automated PR might be interesting?

Not only that but many of the scripts need to be run periodically. So it could be helpful!

mattpaz:
> The two JSON files in the alternate_formats directory should be removed. They don't exist on the main branch. They're generated by CI on the gh-pages branch only.

Yes, I should have caught that.  Just updated to remove the accordingly.

> Yeah that would be great. Sometimes legislators don't have opensecrets IDs when they are sworn in, usually after special elections. And in other cases I may have just been too busy to look it up when adding the legislator.

Makes sense. No worries.  Once this gets committed, I may circle back to this myself.  Which do you think would be more valuable -- work on #886 or work on gaps in opensecrets ids?

>> I wonder if a github action that would run wikidata_update.py on cron and create an automated PR might be interesting?

> Not only that but many of the scripts need to be run periodically. So it could be helpful!

Glad to hear I was on the right track.  If I get time, I might circle back to this and see if I can chip in on that.

JoshData:
Thanks.

> work on #886 or work on gaps in opensecrets ids?

I'm ambivalent. 🤷🏻‍♂️ It's unclear if either are impactful at the moment. But always nice to have complete data.

PR #894, opened 2023-09-29
Sen. Feinstein died

JoshData:
I put today as her term end date since it seems likely she died today, given that it's only breaking now, and not yesterday, but the date hasn't been reported yet as far as I've seen.


dwillis:
@JoshData NYT says she died on Thursday night, fwiw: https://www.nytimes.com/2023/09/29/us/politics/dianne-feinstein-dead-senate.html

JoshData:
Thanks! Updated.

dwillis:
RIP

PR #895, opened 2023-10-03
End Rep. McCarthy's leadership_role for Speaker of the House

JoshData:
Following adoption of H. Res. 757 vacating the office.

I think it will be https://clerk.house.gov/evs/2023/roll519.xml when it is posted.


mlissner:
How the mighty rise and fall.

PR #898, opened 2023-10-15
Rep. Cicilline resigned May 31 and Rep. Stewart resigned Sept 15

JoshData:
Incredibly, we missed these for so long.


dwillis:
Oops!

PR #909, opened 2024-01-04
Kevin McCarthy resigned effective Dec. 31.

dwillis:
https://www.congress.gov/congressional-record/volume-169/issue-214/house-section/article/H7023-6


JoshData:
Thanks!

PR #912, opened 2024-01-26
Updated offices for ben ray lujan

nickoneill:
Office data via: https://www.lujan.senate.gov


nickoneill:
closing in favor of #945

PR #914, opened 2024-02-29
Added all missing opensecrets ids

admarwick:
Except for Laphonza Butler, appointed CA Senator, who has no OpenSecrets id because she
has never raised money.

Closes #913

PR #941, opened 2024-09-03
Update Opensecrets ID for Kean

overtonwindow:
Seems to be wrong? I see this 

https://www.opensecrets.org/members-of-congress/thomas-h-kean-jr/summary?cid=N00027502&cycle=2022


JoshData:
It came from #914. Not sure if OpenSecrets may have changed the ID since then. But seems you're right. Thanks!

PR #943, opened 2024-09-03
Add pictorial ids, script for 118th Congress

### LINKED ISSUE #942: Would this project accept PR with ids for GPO's Pictorial API
I am working on updating a script in unitedstates/images, and it involves matching legislators with their [GPO Pictorial API ids](https://pictorialapi.gpo.gov/api/GuideMember/GetMembers/118). (Unfortunately that API does not include Bioguide IDs.) There are some tricky cases, and so I think the join will require a manual lookup for some legislators with common names. That makes me think the lookup might work well in this repo instead.

Would this repo accept a PR with these ids, and a script to help generate them automatically?

My WIP script is [here](https://github.com/kevinschaul/us-images/blob/f42451ae5baf3159d393ed7b954169120bc33fb9/scripts/gpo_member_photos.py#L77). Happy to transfer it to this repo if you all think that makes sense.

kevinschaul:
Adds ids mapping to GPO's Pictorial Member Guide
pictorialapi.gpo.gov

This PR includes ids for all members of 118th Congress. It may also work for historical files -- at least going back to 110 -- but I have not included that work here, as it requires manual fixes and I imagine most interest is in current membership.

Closes #942


JoshData:
Sorry there is a conflict now.

kevinschaul:
@JoshData Should be all good now

JoshData:
TY!

PR #945, opened 2024-09-07
updated offices for... almost everyone

nickoneill:
Hi again, 5 Calls uses legislator data in this repo and has an interest in keeping district office info updated as alternative phone numbers for constituents to call. I've tried manually keeping this info up-to-date when users tell us it's outdated but it's a pain and doesn't scale. The mechanical turk tool is nice but I wanted something with a faster turnaround time. I experimented with automating this via LLMs and it turned out surprisingly well.

This represents 783 new offices, 406 old offices removed, and added 162 new legislators which had no offices previously.

This data is generated from my new tool, [office-finder](https://github.com/nickoneill/office-finder), which extracts office data from website text for all house and senate websites in about 10 minutes for about $0.20 in openai credits, so anyone can run it relatively frequently. You can read more there, but there's a scrape command as well as a command to update the `legislator-district-offices.yaml` file with the new data.

I tried to be really strict with the existing format of the offices yaml file and follow the validator.py guidelines.

Some notes:
* **I manually verified every single change made to the yaml**, this is not a process we should rely on LLMs to do entirely on their own (that said, I made adjustments to offices for formatting, it still requires some human intervention)
* I retained existing office locations in their current order if they had not changed
* New offices are appended to the end of each list in order by zip code
* The order of fax and phone properties was inconsistent, this will always produce yaml with phone first, then fax, even for unchanged offices, as the validator script desires
* I ran the geocoder tool in this repo after updating the office list for new lat/longs
* Previously the validator script had ~900 lines of warning output, now it has ~300, most of which is geocoding errors (the city name issues in `SANITY_CHECK_EXEMPTIONS` need more entries in the geocoding script) and missing phones when the office is an address only.

I know this is really hard to review a big change like this! Please let me know how I can assist in getting this data merged.

### DIFF
--- a/legislators-district-offices.yaml
+++ b/legislators-district-offices.yaml
@@ -11,8 +11,8 @@
     zip: '35055'
     latitude: 34.181059
     longitude: -86.840631
-    phone: 256-734-6043
     fax: 256-737-0885
+    phone: 256-734-6043
   - id: A000055-gadsden
     address: 600 Broad St.
     suite: Suite 107
@@ -21,8 +21,8 @@
     zip: '35901'
     latitude: 34.01421000000001
     longitude: -86.006846
-    phone: 256-546-0201
     fax: 256-546-8778
+    phone: 256-546-0201
   - id: A000055-jasper
     address: 1710 Alabama Ave.
     suite: '247'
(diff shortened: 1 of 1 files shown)


dwillis:
@nickoneill @JoshData This looks like a really creative and constructive approach to the problem, and I support merging it.

nickoneill:
Hey @JoshData, what are you thoughts on this approach to updating local office information?

JoshData:
Thanks for the PR! Any method that is manually verified is fine with me.

PR #950, opened 2024-11-08
2024 special election results, except Schiff

JoshData:
Four special election results are entered: Ricketts (NE) (succeeding himself), Schiff (CA) (succeeding Butler, who is moved to legislators-historical), Wied (WI), and Carter (TX).

Some information is not yet available:

* The term start dates. My custom for special elections has been to follow the dates listed in bioguide rather than searching for swearing-in dates which might be different. 
* The end dates for Ricketts, Schiff's House term, and Butler also need to be set accordingly.
* Bioguide IDs and birthdays for Weid and Carter.


JoshData:
I updated the PR to drop Schiff's election since it is not yet reflected as a House vacancy (although he didn't vote in today's votes and his term has probably ended by Constitutional requirement since he is an elected senator), or in bioguide, and it is not yet reflected in the Senate (Butler voted today and hasn't left office yet).

And I filled in missing data. So this PR is now ready for review.

JoshData:
Added Gaetz's resignation.

PR #952, opened 2024-11-13
[DRAFT] Special election of Schiff to Senate

JoshData:
Sen. Schiff (CA) succeeding Butler, who is moved to legislators-historical, and Schiff's House term is ended.

The end/start dates are unknown.


JoshData:
Ok so there's an interesting twist with this one. Schiff being sworn-in as a senator today (I didn't check that it happened but it was planned to) is actually by appointment of the governor, and he will be sworn-in again (or ish?) when the special election results are certified, per https://bsky.app/profile/jamiedupree.bsky.social/post/3lcveqdsfik2c. So the `how: special-election` field that I added to his term here isn't quite right. But I think maybe we can overlook this rather than having an extremely short term entry added for this time.

PR #961, opened 2025-01-05
Ran senate_contacts.py, small contact form processing update

nickoneill:
I updated senate_contacts.py to better handle some duplicate contact form URLs that are in the cfm xml and then ran it against the current list.


JoshData:
TY!

PR #962, opened 2025-01-05
Ran house_contacts.py for 2025

nickoneill:
Ran the existing house_contacts.py processor for new addresses and DC phone numbers.


JoshData:
TY!

PR #971, opened 2025-01-20
Run committee membership script

barapa:


JoshData:
The alternate_formats files should not be included in the PR. Thanks.

barapa:
> The alternate_formats files should not be included in the PR. Thanks.

removed from the PR

PR #979, opened 2025-01-29
Removing Michael Waltz from legislators-current and legislators-social-media

amccartney:
First time contributor so let me know if I'm doing this right!

Waltz resigned effective Jan. 20 

https://pressgallery.house.gov/member-data/casualty-list


dwillis:
@amccartney thanks! when a legislator leaves Congress, we move his or her record to the legislators-historical.yaml file (and just delete the social accounts from the social file). Can you update your PR to reflect that?

amccartney:
Very helpful checks

PR #985, opened 2025-02-05
Add contact_form for Schiff's latest term

hartsick:
Noticed this missing, as it broke some of our contact forms. Please let me know if you need anything else! Thanks!


hartsick:
seems like this might not actually be needed since form definition is correct? closing

PR #992, opened 2025-02-21
Add Fax Numbers

putorti:
I'm trying to improve fax coverage; Ed Markey is for sure correct as I pulled it right off of his Senate site, the others I trust the judgement of the core team whether or not to merge. They are indeed fax numbers, but the source is from 2022.


JoshData:
It's ok by me, but if there's no current source it may make sense to drop it.

Thanks.

Issue #1000, opened 2025-04-04
Add: Reps. Jimmy Patronis and Randy Fine


amccartney:
Both were sworn into office after winning their special elections earlier this week 

https://clerk.house.gov/members/P000622
https://clerk.house.gov/members/F000484

PR #1001, opened 2025-04-04
Adding Patronis and Fine

amccartney:
Adding the new reps from Florida to the data, per this issue https://github.com/unitedstates/congress-legislators/issues/1000

### DIFF
--- a/committee-membership-current.yaml
+++ b/committee-membership-current.yaml
@@ -497,18 +497,774 @@ SSAP:
   party: minority
   rank: 14
   bioguide: O000174
-SSAP01: []
-SSAP16: []
-SSAP02: []
-SSAP22: []
-SSAP23: []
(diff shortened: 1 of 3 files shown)


amccartney:
Sorry for delay on this, there aren't yet Govtrack IDs for Patronis and Fine, which seems to be causing the validator to fail. Will finish when those IDs exist.

JoshData:
That's a bit of a chicken-and-egg problem because this is where GovTrack IDs are defined! I do the change here first before loading it into GovTrack. So you can actually just assign the next available number. You can look at the last legislator in legislators-current's govtrack ID and add one, and just double check it's not already in use in legislators-current and legislators-historical (the tests will also check).

JoshData:
Thanks!

PR #1007, opened 2025-05-13
Adding opensecrets ids

admarwick:
Added the opensecret ids to records that had none (76 cases), except where opensecrets has apparently not yet assigned an id (5 cases). The latter are are for one Delegate, two appointed Senators, and two who were the winners of special elections in 2025.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -39273,6 +39273,7 @@
 - id:
     bioguide: K000402
     govtrack: 456957
+    opensecrets: N00054366
     wikipedia: Tim Kennedy (politician)
     wikidata: Q7807373
     votesmart: 91670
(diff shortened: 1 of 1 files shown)


admarwick:
Any ideas on how to fix this check error?

khadersyed:
also, some of the existing open secrets ids are incorrect. @admarwick are you able to include those in here as well? for e.g. Nick Lalota's id is N00045921 - https://www.opensecrets.org/members-of-congress/summary?cid=N00045921

admarwick:
@khadersyed 
Thanks for your attention to this PR. I'm not very fluent in github-ology and I can't see why the PR has not yet been accepted. Do you know why? Since it is still open, I guess that presents an opportunity to update it!

You mentioned one case where the opensecrets id is incorrect. Do you know of any other members whose open secrets ids are wrong? Let me know and I'll add another commit.

khadersyed:
@admarwick Actually, now that I look at it, you fixed all the ones I was looking at - thank you and very much appreciate this PR ❤️

PR #1016, opened 2025-09-08
Fix incorrect CSPAN IDs

### LINKED ISSUE #845: Errors in C-SPAN IDs
The legislators-current.yaml database maps legislators to CSPAN IDs

- Julia Brownley as 79783 but it should be 67561
- Steven Horsford as 79614 but it should be 67777
- Raul Ruiz as 79727 but it should be 67721
- Gerry Connolly as 1015936 but it should be 64005

legislators-historical.yaml has the error
- Joe Kennedy III as 79948 but it should be 67286

Unsure how these errors got introduced.

zkornbluth:
Fixes #845. Verified that the CSPAN IDs for the five legislators were incorrect and replaced them with the correct ones.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -19760,7 +19760,7 @@
     govtrack: 412516
     bioguide: B001285
     thomas: '02106'
-    cspan: 79783
+    cspan: 67561
     votesmart: 59904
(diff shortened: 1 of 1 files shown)


JoshData:
TY!

Issue #1020, opened 2025-09-25
FEC IDs Not Updated


annikamore11:
Is there a lag in which the fec ids are updated? Working on a project where I need to merge bioguide ids with fec data so im merging by the fec ids. However, it seems like some are outdated. 

For instance Sen. John Curtis (UT) has a id$fec = H8UT03238, but has since been reassigned to the new senator candidate id of S4UT00282. Seems like this is true for many of the congress members.

PR #1021, opened 2025-09-25
Update FEC IDs for current legislators

annikamore11:
Added missing fec ids for current legislators, as well as added new fec ids for members who switched positions and got a new id.


JoshData:
Closes #1020

PR #1056, opened 2026-07-26
Add nickname 'Jim' for James E. Risch

bravely:
This pull request makes a small update to the `legislators-current.yaml` file by adding a nickname for Senator James E. Risch. The change adds the `nickname` field with the value "Jim", which is a name he commonly goes by.

### DIFF
--- a/legislators-current.yaml
+++ b/legislators-current.yaml
@@ -1291,6 +1291,7 @@
     first: James
     middle: E.
     last: Risch
+    nickname: Jim
     official_full: James E. Risch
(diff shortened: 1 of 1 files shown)


JoshData:
I don't think we need a nickname for common short forms of first names. In that case, most legislators would probably get a nickname. And I think this would negatively impact how Risch appears on GovTrack because seeing `James "Jim"` everywhere is probably overkill.

eliotpearson:
Should this issue be closed?

